Issue #24, opened 2021-02-25
ADF15 reading


marcosertoli:
Please check implementation of get_adf15 in ADAS class. Also added ADF15 general datatypes, but meaningful dictionary still needs to be implemented.

cmacmackin:
See PR #26.

PR #26, opened 2021-02-25
WIP: Added to adas class new adf15 reading capability

cmacmackin:
I thought a pull request would be more appropriate for this than an issue. From @marcosertoli on #24:

> Please check implementation of get_adf15 in ADAS class. Also added ADF15 general datatypes, but meaningful dictionary still needs to be implemented.

I have marked the PR as a Work in Progress (WIP), to indicate that it's not yet ready to be merged.

### DIFF
--- a/indica/readers/adas.py
+++ b/indica/readers/adas.py
@@ -190,34 +190,30 @@ class ADASReader(BaseIO):
             charge_state = header[1][4:6].lower()
             assert charge_state==charge
 
+            # Read first section header to build arrays outside of reading loop
+            section_header = f.readline().strip()
+            m = re.search(r"(\d+.\d)\s+\S+\s+(\d+)\s+(\d+).+TYPE\s=\s(\S+).+ISEL.+=\s+(\d+)", section_header, re.I)
+            assert isinstance(m, re.Match)
+
+            nd = int(m.group(2))
+            nt = int(m.group(3))
+            data = np.empty((ntrans, nt, nd))
+            ttype = [""] * ntrans
+            tindex = np.empty(ntrans)
+            wavelength = np.empty(ntrans)
+
             # Read Photon Emissivity Coefficient rates
             for i in range(ntrans):
-                section_header = f.readline()
-                m = re.search(r"ISEL =\s*(\d+)", section_header, re.I)
+                if i>0:
+                    section_header = f.readline().strip()
+                m = re.search(r"(\d+.\d)\s+\S+\s+(\d+)\s+(\d+).+TYPE\s=\s(\S+).+ISEL.+=\s+(\d+)", section_header, re.I)
                 assert isinstance(m, re.Match)
-                assert int(m.group(1))-1 == i
-
-                m = re.search(r"TYPE =\s*(\S+)", section_header, re.I)
-                t_type = m.group(1)
-
-                section_header = section_header.split()
-                if i==0:
-                    nd = int(section_header[2])
-                    nt = int(section_header[3])
-
-                    densities = np.fromfile(f, float, nd, " ")
-                    temperatures = np.fromfile(f, float, nt, " ")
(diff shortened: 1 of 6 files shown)


indica/readers/adas.py
```diff
@@ -148,6 +150,119 @@ def get_adf11(self, quantity: str, element: str, year: str) -> DataArray:
...
+                if i==0:
+                    nd = int(section_header[2])
+                    nt = int(section_header[3])
+
+                    densities = np.fromfile(f, float, nd, " ")
+                    temperatures = np.fromfile(f, float, nt, " ")
+                    data = np.empty((ntrans, nt, nd))
+                    ttype = [""]*ntrans
+                    isel = np.empty(ntrans)
+                    wlng = np.empty(ntrans)
```
cmacmackin, line 216: These should be declared before you enter the loop.

indica/readers/adas.py
```diff
@@ -148,6 +150,119 @@ def get_adf11(self, quantity: str, element: str, year: str) -> DataArray:
...
+
+                m = re.search(r"TYPE =\s*(\S+)", section_header, re.I)
+                pec_type = m.group(1)
+
+                section_header = section_header.split()
+                wavelength = float(section_header[0]) / 10. # (nm)
+
+                if i==0:
+                    nd = int(section_header[2])
+                    nt = int(section_header[3])
```
cmacmackin, line 209: Perhaps parse these before entering the loop? that way you could declare data outside of the loop as well and avoid the need for this if/else block.

cmacmackin (commented):
I've requested some stylistic changes. The most important ones are making names more descriptive.

As an aside, I remember you said something about deleting some coordinates so that calculations were possible. I'm not sure how you were doing this, but I'd recommend you use the [`DataArray.drop_vars`](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.drop_vars.html#xarray.DataArray.drop_vars) method. This will create a new data array without the specified coordinates while still retaining the old copy which does have the coordinates, in case they're needed again in future.

indica/readers/adas.py
```diff
@@ -148,6 +150,119 @@ def get_adf11(self, quantity: str, element: str, year: str) -> DataArray:
...
+                assert isinstance(m, re.Match)
+                transition.append(m.group(0))
+
+        gen_type = ADF15_GENERAL_DATATYPES[quantity]
+        spec_type = element #ORDERED_ELEMENTS[z]
+        name = f"log_{spec_type}_{gen_type}"
+        attrs = {
+            "datatype": (gen_type, spec_type),
+            "provenance": self.create_provenance(filename, now),
+            "wlng": wlng,
```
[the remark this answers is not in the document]
marcosertoli: Changed variable names to something more meaningful.
Made wavelength, transition and type coordinates attached to index dimension.

indica/readers/adas.py
```diff
@@ -148,6 +150,119 @@ def get_adf11(self, quantity: str, element: str, year: str) -> DataArray:
...
+            "type": ttype,
+            "transition":transition,
+        }
+
+        print("\n {} ADF15: add info on transition (end of file) \n".format(self.__class__.__name__))
+
+        return DataArray(
+            data * 10**6,
+            coords=[
+                ("isel", isel),
```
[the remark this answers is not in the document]
marcosertoli: Changed to "index".

indica/readers/adas.py
marcosertoli: Moved them to outside of loop

indica/readers/adas.py
marcosertoli: Done

marcosertoli: Done

marcosertoli:
Added adf15 reading capability, changed log_electron... coordinates to log10_electron as in the adf11 files.

PR #27, opened 2021-03-15
Generalize

marcosertoli:
Variables referring to DDA have been generalised to INSTRUMENT in order to generalise for machines other than JET.

Documentation updated accordingly. 
Other variations to the documentation:
- Code name: generalised to Integrated DignostiC Analysis since the composition calculation is one (the first) of the possible applications
- Concept and Workflow: since it gives the details of the composition analysis only, title has been changed to Plasma Composition and the sections explaining the old WSX code have been deleted (original.rst file has been kept as is). Once the new composition analysis workflow has been completed, this section can be expanded.

Still TODO: Input data details / List of data to read: this section must be updated to generalise for any machine and to explain how the data dictionary should be build for each diagnostic (Issue added)

PR #29, opened 2021-03-15
Adas

marcosertoli:
Changed coordinates of ADAS rates from log10_electron.. to electron. This enables broadcasting and simplifies dramatically working with data in the code.

The SCD and ACD rates have been left in log10.


elitherl (commented):
Changes look good and comprehensive, just fixed a couple of typos/formatting errors.

Tests currently failing on `test_adas_reader.py`, error printout:

```
>       assert_allclose(data, result, atol=1e-5)
E       AssertionError: Left and right DataArray objects are not close
E       
E       Differing coordinates:
E       L * electron_density      (electron_density) float64 9.0 9.5 10.0
E       R * electron_density      (electron_density) float64 1e+09 3.162e+09 1e+10
E       L * electron_temperature  (electron_temperature) float64 0.0 0.25 0.5 0.75 1.0
E       R * electron_temperature  (electron_temperature) float64 1.0 1.778 ... 10.0
```

Looks like a mixup between log values in one array and normal space values in the other

cmacmackin:
Fixed the issue with type-checking which was causing pipelines to fail. The problem was that `re.search()` returns _either_ an `re.Match` object or `None`. The type-checker was complaining that the `re.Match.group` method was called on what might have been a `Nonetype` object. This was fixed by adding an assertion check that there was a successful match.

cmacmackin:
As you can see, the linting test is still failing. One issue (probably easy to fix) is you have a line in `snippets/test_ti_measurement.py` that is too long. One of your functions in this file has a [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) that is too high, so you should look at splitting parts of it out into a separate function(s) which it will call. This issue also applies to the ADF15 reader.

cmacmackin:
You have quite a lot of changes present in this PR which do not match the description. This is not good style. Keep your PRs on topic. I see they were orginaly in #27 and you closed that because they'd already been merged into this branch. You should still have approved that PR and merged it into master. Then the changes would no longer show up in the "files changed" display for this PR and you would not have to review them again.

This sort of discipline keeps things clearer and makes PRs easier to manage. It is also helpful for people looking back in future, as it will ensure comments on changes are in the PR which actually introduced them.

cmacmackin:
I've reduced the cyclomatic complexity of the ADF15 reader. This involved reducing the number of loops, if/else blocks, etc. I combined a number of regular expressions to do this. There are no tests provided so I can not confirm that the functionality is exactly the same as before. I may have introduced a bug by mistake. There were one or two things I saw in regular expressions which may have been errors, but not knowing the ins and outs of this ADAS format I couldn't be confident. As such, I did not change the behaviour.

There are a couple of serious concerns which I do have, however. First is the block of code

```python
        if nd == nt:
            coords = [
                ("index", tindex),
                ("electron_density", densities * 10 ** 6),  # m**-3
                ("electron_temperature", temperatures),  # eV
            ]
        else:
            coords = [
                ("index", tindex),
                ("electron_temperature", temperatures),  # eV
                ("electron_density", densities * 10 ** 6),  # m**-3
            ]
```

Why on earth does the order of the axes change if there are the same number of temperature and density data points? That seems bizarre.

I also saw that transition types were being assigned as additional coordinates to the DataArray. This is not an appropriate place to store them, as the values are not all unique (which one would expect of a coordinate). I've moved them over to be attributes (metadata). However, I was wondering whether you'd expect the value of this to change with the index in a single ADF15 file. The file I downloaded to look at as an example (pec40#w_ca#w64.dat) has them all the same. If this is always the case then there is no need to accumulate a list of all of these values and you can just assign a single one as metadata.

marcosertoli:
I had to modify the original function we discussed some time ago in order to account for all the different file formats.

There are changes in the configuration at the end of the file (see ne_pju and c_bnd), differences at the beginning and in the headers (the first two vs pec40 files) and differences in path (transport). This complicates everything exaggeratedly, and I haven't found a way of understanding if there is a logic (apart from the PEC40 and the transport.

If someone can talk to Martin about this (I can't keep on bugging him), if there's a simple logic, I was aiming at having different functions for the different file types, which would have decreased the complexity dramatically.

If this is possible, the data itself is stored in the same manner.

One last point on the coordinates:

        if nd == nt:
            coords = [
                ("index", tindex),
                ("electron_density", densities * 10 ** 6),  # m**-3
                ("electron_temperature", temperatures),  # eV
            ]
        else:
            coords = [
                ("index", tindex),
                ("electron_temperature", temperatures),  # eV
                ("electron_density", densities * 10 ** 6),  # m**-3
            ]


I have no idea why, but it's the only way it works. If you have other solutions, I'd be extremely happy not to have to write this if else statement.

The reason to assign the other coordinates and not as attributes is that one can swap coordinates later on when either interpolating or choosing transitions (with e.g. dataarray.sel()). Some files have 2 or more transitions for the same wavelength (e.g. recombination, excitation & charge exchange). You have to know which is which and be able to select later. I don't think one can you do this with attributes, while coordinates give a good way of selecting, e.g.:


emiss = atomdat["pec"].sel(type="excit") * fz.sel(ion_charges=self.charge)
emiss += atomdat["pec"].sel(type="recom") * fz.sel(ion_charges=self.charge+1)

And, depending on what one is looking for, one will use sel on transition, or type, or wavelength. So we need them all. Can you do any of this using attributes?

So I decided to use index as main coordinate (it's unique for each transition), then have transition configuration and wavelength (both can have duplicates for two or more transitions) and transition type (excitation, recombination or charge exchange).

I know I haven't written it as a RSE would have, but there is logic behind it. Next time, we should discuss before modifying the code further so I can explain the reasons behind the choices made. Then we might converge on something that is RSE-compliant, but meets the needs of the user.

Marco

cmacmackin:
>I have no idea why, but it's the only way it works. If you have other solutions, I'd be extremely happy not to have to write this if else statement.

How does it fail if you don't do it this way? 

>So I decided to use index as main coordinate (it's unique for each transition), then have transition configuration and wavelength (both can have duplicates for two or more transitions) and transition type (excitation, recombination or charge exchange).

OK, I guess that makes sense. I only changed one line so it will be easy to revert. However [multi-level indexing](http://xarray.pydata.org/en/stable/indexing.html#multi-level-indexing) might be a better approach.

marcosertoli:
How does it fail if you don't do it this way?
- it swaps dimensions and I get Ne with Te values, and Te with Ne values...

Multi-level index sounds like a good option. I'll try it out next week and see if it makes more sense. It probably avoids the line of code to swap dimensions.

Marco

cmacmackin:
That is very, very odd. I might try digging into that a bit more, if I have a moment. Could you tell me a dataset where this happens?

marcosertoli:
I agree! I think it's one of the attached, maybe the transport one, but can't check at the moment. It happens only if ne and te have the same number of points. If it's not one of the attached I'll send it over later.

marcosertoli:
It’s the transport file I attached previously.

From: Chris MacMackin ***@***.***>
Date: Friday, 26 March 2021 at 20:24
To: ukaea/Indica ***@***.***>
Cc: marcosertoli ***@***.***>, Author ***@***.***>
Subject: Re: [ukaea/Indica] Adas (#29)

That is very, very odd. I might try digging into that a bit more, if I have a moment. Could you tell me a dataset where this happens?

marcosertoli:
I've tried the adas reader and get an assertion error in line 203.

@Litherland-Smith, ***@***.***> could you please check with Martin O'Mullane if there is a logic behind the files? I'm sure there is because ADAS will have the same "problems" we have, with a special reader for each file with different formatting. If I have that info, I can simplify the reading complexity using separate functions for the different types. @MacMackin, ***@***.***> do you agree?

Marco

cmacmackin:
Yes, that sounds sensible Marco.

I've identified the issue which would be causing this particular bug and pushed the fix.

marcosertoli:
Will do for next time Chris. I'll wait for Ed's input on the file types first after he has contacted Martin. No use we continue to jump back and forth when there might be a better way with that info in our hands.

Marco

elitherl:
I spotted another issue with the regex at line 262 - really an issue in the while loop at 270 though. If it doesn't match `transition_re` it will stay in the loop indefinitely (doesn't seem to stop at end of file), and it doesn't seem to be matching any of the files I've tried. Haven't fixed the regex yet but I've put some quick checks in place in the 3 places with a while loops to prevent infinite line reading.

@marcosertoli I'll have a chat with Martin and let you know what I find out.

marcosertoli:
I'd spotted that, but thought it wasn't worth going further on that until we have info from Martin :-)!

I've reverted locally to my version just to keep on working for the time being.

As soon as Martin gives us the info, I take it on myself to modify it accordingly if you all agree.

Marco

cmacmackin:
We need to resolve the cyclomatic complexity issue on `master` as it is blocking other branches from merging. People are also turning off the pre-commit hooks because of it, which means that even more code is being committed that fails linting. @marcosertoli , you need to provide some tests for ADF15 so that we can get the reduced-complexity version of the read-routine working and merged in.

marcosertoli:
I can't modify the code until Ed comes back to me with file info from Martin.
You can revert to before the adf15 reading function was introduced if this a problem. I have a local version (different file name) that I can the push to a new branch once I have all the info. Someone can then help me write a test for it before merging into master.

Marco

cmacmackin:
I've removed the ADF15 reader from master. This will introduce conflicts when it comes time to merge in this PR, which will need to be resolved manually.

indica/readers/adas.py
```diff
@@ -183,7 +186,175 @@ def get_adf15(
...
+                tindex[i] = i + 1
+                ttype.append(m.group(4))
+                wavelength[i] = float(m.group(1))  # (Angstroms)
+                densities = np.fromfile(f, float, nd, " ")
+                temperatures = np.fromfile(f, float, nt, " ")
+                data[i, ...] = np.fromfile(f, float, nd * nt, " ").reshape((nt, nd))
+                line = f.readline().strip().lower()
+
+            # Read Transition information from end of file
+            file_end_re = re.compile(r"c\s+[isel].+\s+[transition].+\s+[type]")
```
elitherl, line 296: Regex is matching a blank line for the Ar10 test case, causes the transition_match later to fail therefore all tests for Ar10 are failing

tests/unit/readers/test_adf15_reader.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def test_false_input(self):
+        """Checks reading with wrong inputs raises errors"""
+        element = self.element + ";"
+        try:
+            _ = self.reader.get_adf15(
+                element, self.charge, self.file_type, year=self.year
+            )
+        except Exception as e:
+            assert e
```
elitherl, line 39: Can use `.assertRaises` method to check that an `Exception` is raised (as well as checking which to make sure it's the right error you're catching!)

e.g. could replace `try...except` block with: `self.assertRaises(ValueError, self.reader.get_adf15(...))`

indica/readers/adas.py
```diff
@@ -183,7 +186,175 @@ def get_adf15(
...
+            )
+            header_re = re.compile(match)
+            m = header_re.search(transition_line)
+            if not m:
+                transition_type = "n_levels"
+                match = r"c\s+(\d+.)\s+(\d+.\d+)\s+([n]\=.\d+.-.[n]\=.\d+)"
+                header_re = re.compile(match)
+                m = header_re.search(transition_line)
+                if not m:
+                    raise ValueError(f"Unknown transition formatting ({identifier}).")
```
elitherl, line 224: Perhaps worth including the line that failed to match in the exception message to make errors quicker to debug? Up to you if you want to keep `identifier` as well or replace it

tests/unit/readers/test_adf15_reader.py
marcosertoli: Tried this but doesn't work.

indica/readers/adas.py
marcosertoli: Fixed this: it was down to the block **if not line:** which was making it crash.

Still need to set a working EOF block in these while loops. If you have recommendations, I'm open to suggestions!

indica/readers/adas.py
marcosertoli: Seems like this requires the **readlines** method, so either we change to readlines or I'm not sure how this can be done.
It's anyway not too complicated to debug this code, so I have no problems in leaving it as is.

tests/unit/readers/test_adf15_reader.py
gadgil48: There is a way to get `self.assertRaises()` working but it requires creating another class that inherits from `unittest.TestCase` and defining functions inside that and then calling those functions instead of `try...except`, a bit more work than is necessary I think, especially if this works. See `tests/unit/operators/test_fractional_abundance.py` on my issue #42 branch (it's right at the start).

indica/readers/adas.py
elitherl: The `transition_line` argument is the line currently being matched so could just include that. Not desparate though so I won't hold up the merge because of it

tests/unit/readers/test_adf15_reader.py
elitherl: I agree with @gadgil48, expected it to work out of the box but not worth the effort to rewrite since this does the job anyway

cmacmackin: Why is `unittest.TestCase` being used here? Everywhere else we use pytest, which offers the [raises](https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises) context manager.

elitherl (approved):

Issue #10, opened 2021-01-26
Finish implementing Equilibrium methods


cmacmackin:
Currently the methods to get the total magnetic field strength and the volume enclosed by a flux surface (and the inverse of the latter) are not implemented.

PR #34, opened 2021-04-23
Issue #10 : Gadgil48/10 equilibrium methods

gadgil48:
Currently the `enclosed_volume(...)` is implemented as well as a rudimentary unit test for it. Additionally `Btot(...)` is also partially implemented but work on the toroidal magnetic field and the coordinate systems inside the function remains.

### DIFF
--- a/indica/equilibrium.py
+++ b/indica/equilibrium.py
@@ -174,28 +174,26 @@ class Equilibrium(AbstractEquilibrium):
 
         # Mock of sess.prov needed as a fix for a recurring error
         # (probably also a future issue)
-        sess.prov = MagicMock()
         self.prov_id = session.hash_vals(
             **equilibrium_data, R_offset=self.R_offset, z_offset=self.z_offset
         )
-        # self.provenance = sess.prov.entity(
-        #     self.prov_id,
-        #     {
-        #         prov.PROV_TYPE: "Equilibrium",
-        #         "R_offset": self.R_offset,
-        #         "z_offset": self.z_offset,
-        #     },
-        # )
-        # sess.prov.generation(
-        #     self.provenance, sess.session, time=datetime.datetime.now()
-        # )
-        # sess.prov.attribution(self.provenance, sess.agent)
-        # for val in equilibrium_data.values():
-        #     if "provenance" in val.attrs:
-        #         self.provenance.wasDerivedFrom(val.attrs["provenance"])
-        # if T_e and "provenance" in T_e.attrs:
-        #     self.provenance.wasDerivedFrom(T_e.attrs["provenance"])
-        # sess.prov = MagicMock()
+        self.provenance = sess.prov.entity(
+            self.prov_id,
+            {
+                prov.PROV_TYPE: "Equilibrium",
+                "R_offset": self.R_offset,
+                "z_offset": self.z_offset,
+            },
+        )
+        sess.prov.generation(
+            self.provenance, sess.session, time=datetime.datetime.now()
+        )
+        sess.prov.attribution(self.provenance, sess.agent)
+        for val in equilibrium_data.values():
+            if "provenance" in val.attrs:
+                self.provenance.wasDerivedFrom(val.attrs["provenance"])
+        if T_e and "provenance" in T_e.attrs:
+            self.provenance.wasDerivedFrom(T_e.attrs["provenance"])
 
     def Btot(
         self, R: LabeledArray, z: LabeledArray, t: Optional[LabeledArray] = None
@@ -221,11 +219,25 @@ class Equilibrium(AbstractEquilibrium):
             If ``t`` was not specified as an argument, return the time the
             results are given for. Otherwise return the argument.
         """
-        psi = self.psi.interp(
-            t=t,
-            method="linear",
-            assume_sorted=True,
-        )
+        if t is not None:
+            psi = self.psi.interp(
+                t=t,
+                method="linear",
+                assume_sorted=True,
+            )
+
+            f = self.f.interp(
+                t=t,
+                method="linear",
+                assume_sorted=True
+            )
+
+            rho_, theta_, _ = self.flux_coords(R, z, t)
+        else:
+            t = self.rho.coords["t"]
+            psi = self.psi
+            f = self.f
+            rho_, theta_, _ = self.flux_coords(R, z)
 
         dpsi_dR = psi.differentiate("R").indica.interp2d(
             R=R,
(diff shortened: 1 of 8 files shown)


cmacmackin:
I notice that there are a lot of formatting issues being flagged in the linting tests. I take it you've turned off the pre-commit hooks because of the issue introduced in #26? I've made a note on #29 that it needs to be completed and merged as soon as possible to fix that. If that doesn't happen then I'll revert the changes introduced by #26 so we can get linting working for everyone else.

tests/unit/data_strategies.py
```diff
@@ -282,14 +282,14 @@ def data_arrays_from_coords(
         x2_scaled = (x2 - min_val) / (width if width else 1.0)
     else:
         x2_scaled = 0.0
-    tmp = func(t_scaled, x1_scaled, x2_scaled)
+    tmp = func(x1_scaled, x2_scaled, t_scaled)  # t_scaled at the end
```
cmacmackin, line 285: Pretty sure `t_scaled` should be at the start. It is always the first one in `dims`.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,498 @@
...
+from xarray import where
+
+from indica.converters import FluxSurfaceCoordinates
+from indica.converters import TransectCoordinates
+from indica.converters import TrivialTransform
+from indica.converters import LinesOfSightTransform
+from indica.equilibrium import Equilibrium
+
+from scipy.optimize import fmin
+
```
cmacmackin, line 20: You probably don't need to make all of these routines as general as I did for Property Based Testing. You can just have them generate one, simple, type of data. You might find that you don't need `smooth_funcs` at all anymore, for example. (I'll leave a comment with a more concrete example below). It looks like you've already done a lot of the work on this, so if things work then you might as well just leave it as is, but this could save time in future.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,498 @@
...
+    abs_sigma=1e-3,
+    uncertainty=True,
+    max_dropped=0.1,
+    require_dropped=False,
+):
+    general_type = (data_type[0])  # To-do: error-checking
+    specific_type = (data_type[1])  # To-do: error-checking
+
+    x1, x2, t = axes
+    func = noisy_funcs(data, rel_sigma, abs_sigma) \
```
cmacmackin, line 289: I don't think we actually need to test with noisy data, to be honest.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,498 @@
...
+    tmp = func(x1_scaled, x2_scaled, t_scaled)
+    result = xr.DataArray(np.reshape(tmp, shape), coords, dims)
+
+    if isinstance(x1, np.ndarray):
+        flat_x1 = x1.flatten()
+    else:
+        flat_x1 = x1
+    dropped = (
+        [
+            flat_x1[i] for i in dropped_chnls(
```
cmacmackin, line 332: We probably don't need to worry about testing cases with dropped channels, either.

cmacmackin (commented):
Overall things look alright. There are a few stylistic things and some cases where you can probably do things more simply. The latter would mostly be because of you continuing my over-complicated approach. The most important thing to change is how you going about mocking the Session object, on which I've provided some more detailed annotations.

Looking at the CI logs, once you turn linting on again there will be various changes that are required. Some of these can be done automatically. There are some type-errors though, which arise from you using methods which are not guaranteed to be available on all of the allowed types for an argument. You'll need to add some explicit type-guarding in those cases (see [mypy docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#union-types)). 

It also looked like MyPy complained about you dividing a DataArray by an integer at one point. If I remember correctly, this is due to a bug in the typing information in xarray. You'll just have to turn off type-checking on those lines by appending `# type: ignore`.

cmacmackin:
I have removed the code which causes linting failure (#35). If you merge that into this branch and run `poetry run pre-commit run -a` it should do all necessary reformatting and point out any other issues which need to be fixed manually.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,600 @@
...
+
+    """Magnetic field strength at maximum height of the device (whilst inside LCFS)
+    """
+    Total_B, _ = equilib.Btot(max_height_R, max_height_z, time)
+
+    rho_, theta_, t_ = equilib.flux_coords(max_height_R, max_height_z, time)
+    test_nan_B(Total_B, rho_)
+
+    """Magnetic field strength at minimum height of the device (whilst inside LCFS)
+    """
```
cmacmackin, line 578: Use standard comments, with the hash symbol.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,600 @@
...
+    Total_B, _ = equilib.Btot(equilib.psi.coords["R"], equilib.psi.coords["z"], time)
+
+    rho_, theta_, t_ = equilib.flux_coords(
+        equilib.psi.coords["R"], equilib.psi.coords["z"], time
+    )
+    test_nan_B(Total_B, rho_)
+
+    """Magnetic field strength at all locations on the (R, z) grid at all times
+    """
```
cmacmackin, line 594: Use standard comments, with the hash symbol.

tests/unit/test_equilibrium_single.py
```diff
@@ -0,0 +1,600 @@
...
+        variance = max_std * rel_sigma * np.max(np.abs(y)) + max_std * abs_sigma
+        elements = np.random.random_sample(y.shape)
+        elements = (2.0 * elements) - 1.0
+        elements *= variance
+        return y + elements
+
+    return noisy
+
+
+def dropped_chnls(size, max_dropped=0.1):
```
cmacmackin, line 239: Call this `dropped_channels` (i.e., don't abbreviate).

indica/equilibrium.py
```diff
@@ -113,38 +110,42 @@ def __init__(
...
+            #     fluxes = self.rho.interp(
+            #         t=T_e.coords["t"], method="nearest"
+            #     ).indica.interp2d(
+            #         R=T_e.coords["index"] - offset,
+            #         z=T_e.coords["index_z_offset"] - z_shift,
+            #         method="cubic",
+            #         assume_sorted=True,
+            #     )
+            #     offset, accept = offset_picker(offset, T_e, fluxes, best_fits)
+            self.R_offset = R_shift
```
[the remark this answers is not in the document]
gadgil48: This was causing a few problems when giving a valid `T_e` to initialize the Equilibrium class. I'm going to create a separate issue relating to this as I think the problem goes beyond the code I've commented out. The value for `T_e` returned by `equilibrium_dat_and_Te()` is currently `None`, this should be fixed by fixing the problem for the commented out code in the Equilibrium class initialization.

indica/equilibrium.py
```diff
@@ -317,11 +379,47 @@ def enclosed_volume(
...
         """
-        raise NotImplementedError(
-            "{} does not implement an 'enclosed_volume' "
-            "method.".format(self.__class__.__name__)
+        if t is None:
+            t = self.rho.coords["t"]
+
+        Major_radius_axis = self.rmag.interp(
+            t=t,
+            method="linear",
```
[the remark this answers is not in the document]
gadgil48: When the length of the interpolating(or interpolated?) coordinate(`t` in this case) is `<=3` then the "cubic" interpolation fails but "linear" works. I've put a check in now to determine whether to use "linear" or "cubic" depending on the length of `t` where relevant.

indica/equilibrium.py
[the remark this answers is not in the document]
cmacmackin: Definitely something that needs its own issue.

indica/equilibrium.py
```diff
@@ -161,6 +162,8 @@ def __init__(
             np.arctan2(self.zmin - self.zmag, self.Rmin - self.rmag) % (2 * np.pi),
         ]
 
+        # Mock of sess.prov needed as a fix for a recurring error
+        # (probably also a future issue)
```
cmacmackin, line 166: Is this still the case. If so, is it only a problem you've had when running unit tests or have you noticed it in more realistic tests of the code on real data too?

cmacmackin (approved):
Thanks Sanket, this all looks good. I just had one more question. Could you also raise an issue about the problem with `T_e`, then everything should be ready to merge.

indica/equilibrium.py
gadgil48: Sorry, this is a leftover comment from before I put in the fix you suggested (`sess=MagicMock()` keyword argument).

cmacmackin: Happens to everyone.

cmacmackin:
The test failure you got after your last push was due to Hypoethesis happening upon some input for the first time which could produce a failure. As you might have seen, it had nothing to do with your own code. I've fixed the issue for you.

PR #35, opened 2021-04-27
Remove ADF15 reader code

cmacmackin:
This code has too high a cyclomatic complexity and is causing linting
to fail. The issue will be addressed in #29 but until then this code
is being removed from master. Its presence is preventing other PRs
from being merged and causing people to turn off the pre-commit
linting (introducing yet more problems).


cmacmackin:
I've also turned off the running of flake8 on snippets. These often won't be up to the standard of code we expect in the main code-base.

Issue #39, opened 2021-06-14
ValueError on calling R_hfs with a given t


elitherl:
**Describe the bug**
Calling `Equilibrium.R_hfs` with given values for both `rho` and `t` results in a call to `R = rmji.interp(rho_poloidal=rho, method="cubic") - self.R_offset`. As both `rho` and `rmji` have a common coordinate `t`, `interp` attempts 2D interpolation and `method="cubic"`

elitherl:
```python
        #  Snippet from equilibrium.py
        if t is None:
            rmji = self.rmji
            t = self.rmji.coords["t"]
            rho, _ = self.convert_flux_coords(rho, t, kind, "poloidal")
            R = rmji.indica.interp2d(rho_poloidal=rho, method="cubic") - self.R_offset
        else:
            rmji = self.rmji.interp(t=t, method="nearest")
            rho, _ = self.convert_flux_coords(rho, t, kind, "poloidal")
            R = rmji.interp(rho_poloidal=rho, method="cubic") - self.R_offset
```
If `t` isn't given to `R_hfs`, `.indica.interp2d` is used instead. Changing the case where `t` is given to also use this solves the issue

elitherl:
Solved by #40

PR #40, opened 2021-06-14
Change use of .interp to .indica.interp2d

elitherl:
Change interpolating rmji using .interp to using .indica.interp2d in case where t is passed to R_hfs. Fixes issue #39 with nd interpolation not allowing cubic method and is consistent with case where t isn't given


gadgil48 (approved):

gadgil48 (approved):
This looks right now. Good to have the previous version included as well.

gadgil48 (approved):
Thanks for adding this. Good to know that the framework in `test_equilibrium_single.py` for generating equilibrium data is easy enough to understand and use.

PR #42, opened 2021-06-15
Fractional abundance and total radiated power

### LINKED ISSUE #19: Calculate fractional abundance
**Calculate the ionisation balance and radiation loss parameter of specified element**
Given the ionisation and recombination rates, line and recombination radiative power parameters of an element and an electron temperature array, calculate the fractional abundance of each ionisation stage and the cooling factors, total and SXR.  

**I/O of module**
Parameters to the module:
* ionisation and recombination rates, line and recombination radiated power coefficients as read from OpenADAS, function of electron density, temperature and ion charge  (Ne, Te, q)
* electron temperature e.g. from spline-fit of diagnostic data, function of normalised radial coordinate and time (rho, t)

Keywords
* electron density = 5.e19: at typical tokamak densities, the rates are ~constant, so average value can typically be used

Interpolate the rates and radiated power coefficients on the electron temperature profiles (optional on density profiles, otherwise interpolate on default constant value) and calculate the equilibrium fractional abundance f_Z(rho, t, q) and the total radiated power loss. 

Return f_Z(rho, t, q), rad_power(rho, t, q).

**Additional context**
These will be necessary to calculate the mean charge of each ion along the profile, which will then be used to estimate the contributions of each ion/element to Zeff and dilution, as well as for the forward model of any spectroscopic diagnostic (e.g. VUV, KT7/3) measuring a specified ionisation stage or a bundle of them. 
Total/SXR-filtered radiated power will be used to calculate the density of high-Z and medium-Z impurities from SXR and bolometric data.

gadgil48:
Recommended merge order is to merge in the order the issues are listed in "Complete core computational modules" milestone.
Closes #19


elitherl (approved):
Looks good, only comment was on the implementation of `ordered_setup` which has now been changed

marcosertoli:
Similar to area and volume, the fractional abundance and radiated power calculation should be in separate functions:
- FractionalAbundance should NOT have in input PLT, PRB, PRC and output only the fractional abundance
- CoolingFactor should NOT have the fractional abundance in input (in place of SCD, ACD and CCD) and output the cooling factor.

The reason for splitting the two functions is that the fractional abundance can be calculated through other means, e.g. using a code which includes transport coefficients.

Nh/Ne is to be multipled to PRC, not PRB.

The variables N_z_t (and N_z_t0), should be called F_z_t (F_z_t0), since it is the fractional abundance and not the density that is being calculated.

The charge exchange parameters (CCD and Nh for FractionalAbundance, PRC for CoolingFactor) should be optional keywords since they are often either not required for the calculation or not known.

marcosertoli:
Sanket, there are still a few details we have to sort before merging. But they're really small.
- If the file is to contain different operators for atomic data calculation, which I think makes sense, we might as well call it: atomic_data.py or something similar, so that inside it we can have the two (or more) classes e.g. FractionalAbundance and PowerLoss. Otherwise, with the current name the file should contain only the fractional abundance class.
- Nh should be optional also in PowerLoss.
- If F_z_t is None, then the function should return the interpolated cooling functions only. It wouldn't be correct to do otherwise, especially since the fully-stripped will not be living across the whole radius, whether it's Be or W we're talking about.
- All the inputted quantities apart from F_z_t should be != nan, != inf and >0. Why add checks for Nh only? Can you not do the checks for all input quantities following the dictionary you build in the __init__?
- You're also missing checks on F_z_t in the Power class. 
- More on the checks: since they're all the same, can you not use the same functions for both classes putting it outside of the class structure? And since the checks have to be done at initialisation, I don't see the need to make it a class attribute. This would save a lot of space and make the functions available to other operators that will be built in this file.
- You're asking for DataArrays in input, which is fine so we can follow the provenance (which diagnostic this data comes from etc. or, if it's invented, the DataArray attributes will show it). But the output MUST be a DataArray with all the provenance in there! Otherwise no use having the input as a DataArray...
- Still in this line, the provenance of the outputs should include info on which atomic data files have been used, as well as temperatures, densities, etc.

elitherl:
I second those points about checks, I think the common checks are perfect to be abstracted so they can be reused (especially if more operators are to be added that use them).

Only other thing was on coverage: there are a few checks on input parameters being there or not, or that they fit certain restrictions, and should just need a couple more tests to make sure those checks are being hit when expected. Otherwise looks good from my side.

marcosertoli:
One more point: I don't see any difference between inputted and imported data, apart from the dimensions. Might want to call them with a name that is more related to the quantity of interest e.g. atomdat for ACD, SCD, etc. and kinetic for ne, te, etc.
So yes, I'd be for putting all the checks in one function, e.g. your input_check and take everything else away.

PR #43, opened 2021-06-17
Separated calculation of cross-sectional area and now using trapezoidal rule for integration

### LINKED ISSUE #37: Area equilibrium method
**Feature request**
Poloidal cross-sectional area is calculated within the enclosed_volume method of equilibrium.py (variable name = "area").

Make the calculation of "area" a separate method of equilibrium.py and call it inside enclosed volume.

**Additional context**
The "area" is necessary for computation of several plasma quantities so the use of a separate method is necessary.

gadgil48:
Closes #37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ htmlcov/
 .dir-locals.el
 *.pkl
 indica/git_version
+.vscode/
--- a/indica/equilibrium.py
+++ b/indica/equilibrium.py
@@ -2,6 +2,7 @@
 """
 
 import datetime
+import typing
 from typing import cast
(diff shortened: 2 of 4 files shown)


marcosertoli:
A few questions:
- do you get the same results you did previously? 
- How much faster is the new method with respect to the previous?
- What is the minimum theta you need to reproduce the previous result?

marcosertoli (approved):
Looks good Sanket, not more comments from my side

gadgil48:
Folded in a fix for array of rhos not being accepted in `enclosed_volume()`. Closes #44

PR #46, opened 2021-06-30
V2: Separated calculation of cross-sectional area and now using trapezoidal rule for integration

### LINKED ISSUE #37: Area equilibrium method

gadgil48:
Please merge this before all other "Complete core computational modules" milestone issue pull requests.
Closes #37


elitherl:
Something's wrong and I can't work out why, this PR wants to revert/remove the changes merged in from the adas PR but the branch is fully merged with master, and I suppose that's why the coverage has dropped too. Any idea why and how to fix it?

gadgil48:
Spoke with @cmacmackin about the code coverage failing. Seems that any code refactoring which reduces the number of existing covered lines causes coverage to decrease (even if the new added lines are covered). This a known issue with `codecov`, see: https://community.codecov.com/t/project-coverage-decrease-on-deleted-line/2726

For now Chris has disabled the `codecov` check for the project (it's still performed but it's not a requisite for merging anymore).

Going forward this may resolve itself once the requisite number of covered lines that were removed by this branch (and possibly by Marco's adas branch - #29 ) are restored later as more functionality and tests are added.  Keep in mind these would not be the same lines but the same _number_ of lines as that's what `codecov` cares about.

elitherl:
That's interesting with codecov, good to know why that's happening and hopefully there's a fix sometime soon.

Any idea on the reverting changes from #29 though? I tested by branching off master and merging, it's not GitHub reporting wrong and it actually removes the changes that were added

marcosertoli (changes requested):
Sanket, it all look fine to me and I'd be ready to merge, but there are two last details, really small, to change before merging:
- flux_surfaces._convert_to_vol : no need for theta since you give rho. And output is still vol, t, area instead of vol, area, t (same in fake_equilibrium test). This should also return area for consistency with the function. 
- it's worth adding a _convert_to_area for symmetry and since this might be necessary if volume is not needed.
- please put a TODO in the abstract equilibrium since the encosed_volume converter calls a not implemented function to invert from volume to rho. No need to do it now, maybe we will never need it...
Whenever you're done with this I'm ready
 to merge!

cmacmackin:
Why is cross-sectional area being returned by the same methods as enclosed volume at all? Wouldn't it make more sense for those to be in separate methods?

marcosertoli:
Sure, but since cross sectional area is calculated anyway, it's worth returning it. While if you need only area you can ask only for that.

elitherl:
@marcosertoli your review is still requesting changes, did the latest updates address them all? If so, please re-review and approve and we'll be good to merge this pr

elitherl (approved):

Issue #25, opened 2021-02-25
Test PPF reader in CI


cmacmackin:
**Is your feature request related to a problem? Please describe.**
Tests of the PPF reader rely on having sample PPF data saved in a pickle file. However, as the data is proprietary, we may not commit it to the repository. This means it is not available for use on CI and those tests must be skipped, as described in issue #2.

**Describe the solution you'd like**
Anonymise the data sufficiently that we'd be allowed to store it in the repository. We'd need to consult with legal to determine what this would require.

**Describe alternatives you've considered**
Alternatively, we could encrypt the data and [decrypt it within the CI](https://docs.github.com/en/actions/reference/encrypted-secrets#limits-for-secrets). This would be less elegant as it would not allow developers to run these tests. However, depending on what we're told by legal, this might be the only acceptable solution.

PR #50, opened 2021-07-16
Make PPFReader tests runnable on CI

cmacmackin:
PPF data is proprietary, meaning we can not commit it to the repository. This posed a challenge for my approach to testing the PPF reader, as that relied on having a sample of some PPF data in a pickle file (#25). However, I realised that I just need data that is the _right shape_, and the actual values could be overwritten with random numbers. I've modified the script which fetches the data to make that possible and committed a sample of this dummy data to the repository. We will now be able to run these unit tests in CI (which should give a nice boost to code coverage!).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ htmlcov/
 .dir-locals.el
 *.pkl
 indica/git_version
+
+!tests/unit/readers/ppf_samples.pkl
--- a/tests/unit/converters/test_time.py
+++ b/tests/unit/converters/test_time.py
@@ -21,7 +21,10 @@ from indica.converters.time import convert_in_time
 from indica.utilities import coord_array
 from .test_abstract_transform import coordinate_transforms_and_axes
 from ..data_strategies import data_arrays_from_coords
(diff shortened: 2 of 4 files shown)


elitherl:
Sorry it took a little while to get to this, had a look through and changes look good.

My only comment is on `test_get_cyclotron_emissions` (`test_ppf_reader.py`) - running the tests locally I'm getting a Flaky failure due to some variability in test run time, 292.24ms vs 160.87 ms where the deadline is 200ms by default. It hasn't failed on GitHub though, but perhaps worth increasing the deadline so we don't run into this error later on and think something else is causing it?

cmacmackin:
The reason there weren't any such warnings on GitHub is because I configured hypothesis not to check time limits when running on CI. This is partially because it is too difficult to debug; my laptop was faster so things that were within the time limit locally would fail on CI. I've increased the deadline for the test now.

elitherl:
That makes sense, and I agree that's the most sensible way of doing things then. All looks good in that case.

elitherl (approved):

PR #56, opened 2021-08-20
Fix for issue #36

### LINKED ISSUE #36: Fix calculation of self.R_offset in the Equilibrium class initialization for a valid T_e
When a valid `T_e` is given to the `Equilibrium` class for initialization, an error is produced at the calculation of `self.R_offset`. This is eliminated if `T_e = None` is passed to Equilibrium instead. Given the nature of the problem there is likely to be fixes needed for `electron_temperature()` in `test_equilbrium_single.py` therefore for testing of `enclosed volume()` and `Btot()`, `T_e = None` is passed to the Equilibrium class. This will be rectified when the fix for this issue is complete.

gadgil48:
Closes #36

### DIFF
--- a/indica/equilibrium.py
+++ b/indica/equilibrium.py
@@ -2,8 +2,10 @@
 """
 
 import datetime
+from typing import Any
 from typing import cast
 from typing import Dict
+from typing import Hashable
 from typing import Optional
 from typing import Tuple
 
@@ -20,6 +22,7 @@ from .abstract_equilibrium import AbstractEquilibrium
 from .numpy_typing import LabeledArray
 from .offset import interactive_offset_choice
 from .offset import OffsetPicker
+from .operators import SplineFit
 from .utilities import coord_array
 
 _FLUX_TYPES = ["poloidal", "toroidal"]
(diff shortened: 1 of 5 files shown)


indica/equilibrium.py
```diff
@@ -110,42 +114,47 @@ def __init__(
...
+                    z=T_e.coords["index_z_offset"] - z_shift,
+                    method="cubic",
+                    assume_sorted=True,
+                )
+                offset, accept = offset_picker(offset, T_e, fluxes, best_fits)
+
+                # For escaping the while loop while unit testing. Otherwise
+                # with offset_picker = MagicMock(return_value=(0.02, False)),
+                # then this while loop will not end.
+                if unit_testing:
```
elitherl, line 155: Instead of adding a specific `unit_testing`, could replace offset with
```python
offset = MagicMock(side_effect=[(0.02, True), (0.02, False)])
```
and it will iterate through those return tuples per call, going into the `while not accept` after first call but exiting after the second

elitherl (approved):

PR #57, opened 2021-08-26
Fix for issue #55

### LINKED ISSUE #55: Move `input_check` function into abstract `Operator` class
Similar variations on an `input_check` function are used across many subclasses of the abstract `Operator` class, all performing the same role. Moving this to be a method of `Operator` reduces duplicated code and makes the check available for new operator classes being added without remaking the function

gadgil48:
Closes #55

### DIFF
--- a/indica/operators/atomic_data.py
+++ b/indica/operators/atomic_data.py
@@ -1,5 +1,4 @@
 import copy
-from typing import get_args
 from typing import List
 from typing import Optional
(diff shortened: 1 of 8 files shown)


gadgil48:
Code coverage is going wrong again because of the removal of previously covered lines (even though the replacement code is covered). Should be fine to merge if everything else is fine.

elitherl (approved):

PR #67, opened 2021-09-15
Restructure of atomic_data.py

### LINKED ISSUE #62: Modification of FractionalAbundance class in atomic_data.py
**Describe the bug**
We’re currently initializing the class giving Ne, Te, Nh in input. This is wrong. The only inputs necessary in the initialisation are the atomic data. Ne, Te, Nh, F_z_t0 should be given in input when calling the method to calculate the Fz

**Minimal Reproducer**
Move the desired parameters to the __call__. 

Ne, Te compulsory, F_z_t0 and Nh optional, the latter conditional on having given CCD rates in the init. Interpolation of the rates on the inputs shall be performed here, before calculating F_z.

Tau should also be optional (tau=None): 
- if tau == None, calculate for tinf 
- if tau != 0 proceed as in current call

**Additional context**
The reason for this is that we’re defining the ACS, SCD, CCD at the beginning and these will be applied throughout the whole iterative optimisation. When changing Ne, Te, Nh, tau, we don’t want to rebuld the object or perform checks on the inputs, we just want to interpolate the atomic data on Ne, Te, Nh and calculate the output.

gadgil48:
Closes #62

### DIFF
--- a/indica/numpy_typing.py
+++ b/indica/numpy_typing.py
@@ -20,7 +20,7 @@ except ImportError:
     ]
 
 LabeledArray = typing.Union[
-    float, int, xarray.DataArray, xarray.Dataset, xarray.Variable
+    float, int, np.ndarray, xarray.DataArray, xarray.Dataset, xarray.Variable
 ]
 
 OnlyArray = typing.Union[
--- a/indica/operators/atomic_data.py
+++ b/indica/operators/atomic_data.py
@@ -270,7 +270,7 @@ class FractionalAbundance(Operator):
 
         self.interpolation_bounds_check(Ne, Te)
 
-        self.Ne, self.Te = Ne, Te
+        self.Ne, self.Te = Ne, Te  # type: ignore
 
         SCD_spec = self.SCD.indica.interp2d(
             electron_temperature=Te,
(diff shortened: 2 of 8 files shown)


marcosertoli:
The code looks ok, but I'm getting a few inconsistencies when trying it out.

1. we want to initialise the class with all the available atomic data i.e. including CCD also, but we must be allowed to not to include Nh in the call. If this happens Nh should be internally set to a zeros_like with the same shape as the other inputs. Here's the error I get:
`----> 1 fz = example_frac_abundance(input_Ne, input_Te)

~/Work/Python/Indica_github/Indica/indica/operators/atomic_data.py in __call__(self, Ne, Te, Nh, tau, F_z_t0, full_run)
    626             self.interpolate_rates(Ne, Te)
    627 
--> 628             self.calc_ionisation_balance_matrix(Ne, Nh)
    629 
    630             self.calc_F_z_tinf()

~/Work/Python/Indica_github/Indica/indica/operators/atomic_data.py in calc_ionisation_balance_matrix(self, Ne, Nh)
    335             inputted_data["Nh"] = Nh
    336         elif self.CCD is not None:
--> 337             raise ValueError(
    338                 "Nh (Thermal hydrogen density) cannot be None when\
    339                 CCD (effective charge exchange recombination) at initialisation \

ValueError: Nh (Thermal hydrogen density) cannot be None when                CCD (effective charge exchange recombination) at initialisation                 is not None.
` 

2. documentation internal to each method has not been updated to new version of inputs/outputs

3. Minor point: can you remind me why do Ne, Te and Nh have to be DataArrays? If we're not giving a second dimension (e.g. time) to the inputs, then they could be simple arrays, correct? It is important that the outputs are data arrays, but the inputs don't necessarily have to be.

marcosertoli (approved):
I trust you and Ed 100% on the testing, the rest looks good and works fine!

PR #69, opened 2021-09-16
Code name change

### LINKED ISSUE #68: Name change for InDiCA and update .pre-commit-config.yaml
**Is your feature request related to a problem? Please describe.**
Change the naming for InDiCA to better reflect the code's capabilities. Also update the `.pre-commit-config.yaml`

gadgil48:
Closes #68 

Indica was planned to go beyond the original IDL code and it is doing exactly that.

@marcosertoli  has updated the README to reflect this.

@marcosertoli  also had to do an autoupdate to pre-commit because it was giving me errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,4 @@ repos:
     rev: v0.910  # Use the sha / tag you want to point at
     hooks:
     -   id: mypy
+        additional_dependencies: [types-setuptools]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,16 @@
 repos:
 -   repo: https://github.com/sqlalchemyorg/zimports/
-    rev: 0.2.0
+    rev: v0.4.1
     hooks:
     - id: zimports
       name: "Reorganise Import Statements"
 -   repo: https://github.com/psf/black
-    rev: stable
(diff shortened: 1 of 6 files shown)


gadgil48:
Sorry about the new pull request @marcosertoli  but I created an issue for the branch and renamed the branch and the only way to change the name of a remote branch seems to be to delete the old branch and create a new one with a new pull request. For consistency I've changed the pull request description to state that the changes made have been made by you (even though the pull request is now identified as being created by me).

Additionally I'm not sure what happened with `mypy` and the other packages on your local machine that meant they needed to be updated in `.pre-commit-config.yaml` but I've kept them as they were before your changes. This is since the amount of compatibility issues with the frozen versions of all of the dependencies in `poetry.lock` and `pyproject.toml` (`numpy`, `scipy`, etc.) were simply too many to untangle in a straightforward manner. The static analysis still works, and we may need to update _all_ of the packages that are pre-requisites but this will need to be done carefully to ensure that there are no incompatibilities with the code itself and the new packages.

marcosertoli:
Sanket, I completely support your approach.

Please do update all the packages if needed. Interact with Ed and call Chris to have his take on it if needed.

Your current branch has two more files changed. If this is not necessary for this branch, we should revert these files to master and just change readme as I initially intended to do.

PR #70, opened 2021-09-20
Introduction of new dictionary in datatypes.py for combining all elements properties

### LINKED ISSUE #60: Element name, charge and mass
Change all the ELEMENTS dictionaries in datatypes.py to a single dictionary with keys = element symbol ("H", "D", "He3", "W", etc.) and items = (charge, mass, element name).

Adapt all code relating to these quantities.

gadgil48:
Closes #60


indica/operators/impurity_concentration.py
```diff
@@ -76,8 +76,8 @@ def __call__(  # type: ignore
         Parameters
         ----------
         element
-            String specifying the element for which the impurity concentration
-            is desired. It should be given in full lower-case, eg. "beryllium"
+            String specifying the symbol of the element for which the impurity
```
elitherl, line 79: Should this, and it's usages, be changed to use element symbol instead of name for consistency with other uses?

indica/operators/mean_charge.py
```diff
@@ -73,21 +73,21 @@ def __call__(self, FracAbundObj: DataArray, element: str):  # type: ignore
...
             raise ValueError(
                 f"Please input a single valid element from list:\
-                {list(ELEMENTS_BY_ATOMIC_NUMBER.values())}"
+                {list(ELEMENTS.keys())}"
             )
 
-        element_atomic_number = element_atomic_number_tmp[0]
+        element_atomic_number = ELEMENTS[element][0]
 
         ionisation_charges = np.linspace(
```
elitherl, line 86: This could be simplified to just `np.arange(0, element_atomic_number + 1)` looking at the fact it's setting step to `element_atomic_number + 1`. Would also remove the need for `dtype=int` (although it doesn't hurt to keep that just to enforce it)

indica/readers/adas.py
```diff
@@ -133,7 +131,14 @@ def get_adf11(self, quantity: str, element: str, year: str) -> DataArray:
                     date = new_date
                 data[i, ...] = np.fromfile(f, float, nd * nt, " ").reshape((nt, nd))
         gen_type = ADF11_GENERAL_DATATYPES[quantity]
-        spec_type = ORDERED_ELEMENTS[z]
+        spec_type_tmp = [value[2] for value in ELEMENTS.values() if value[0] == z]
+        # This is basically only needed for hydrogen, deuterium and tritium,
+        # where z=1 for all of them. Other elements in the datatypes.py have a
+        # unique atomic number.
```
elitherl, line 137: Could do this find/assert the opposite way to simplify things: find the element based off `spec_type` since that's the unique part and then assert the Z from that element is equal to `z` from the file

indica/operators/impurity_concentration.py
gadgil48: I think I've caught all of the uses of element name and changed them to symbol but if I've missed any please let me know. Do you mean should the change from element name to element symbol have been made at all?

indica/operators/mean_charge.py
gadgil48: You're right, it's simpler but `mypy` still throws an error even with `dtype=int` being specified. Either way I'll change it to use `np.arange`

indica/readers/adas.py
gadgil48: Yep definitely, I'll make the change

indica/readers/ppfreader.py
```diff
@@ -255,7 +255,11 @@ def _get_charge_exchange(
         results["R"] = R.data[0, :]
         results["z"] = z.data[0, :]
         results["length"] = R.data.shape[1]
-        results["element"] = ELEMENTS_BY_MASS[int(round(mass.data[0]))]
+        results["element"] = [
+            value[2]
+            for value in ELEMENTS.values()
+            if value[1] == int(round(mass.data[0]))
+        ][0]
```
elitherl, line 262: Not an issue with this change but isn't getting element name by mass dangerous since mass is non-unique?
CX also defines a dtype "ZQNN" that gives the Z number of the main impurity, could use that to select element? (either instead or as well as mass, since Z and mass will be unique even with H isotopes)

indica/operators/mean_charge.py
```diff
@@ -73,22 +73,17 @@ def __call__(self, FracAbundObj: DataArray, element: str):  # type: ignore
...
+                {list(ELEMENTS.keys())}"
             )
 
-        element_atomic_number = element_atomic_number_tmp[0]
+        element_atomic_number = ELEMENTS[element][0]
 
-        ionisation_charges = np.linspace(
-            0, element_atomic_number, element_atomic_number + 1, dtype=int
-        )
+        ionisation_charges = np.arange(0, element_atomic_number + 1)  # type: ignore
```
elitherl, line 86: Hmm, that's annoying about mypy throwing an error. If we wanted to un-ignore it we could also type hint in place since we know what `np.arange` has to return e.g. `ionisation_charges: np.ndarray = np.arange(...)`? No problem leaving it as ignored either if that doesn't actually gain anything for us though

elitherl (approved):
Looks good. Remaining comments are non-critical and only there if you think they're worth changing. If you have a look and want to leave as-is I'm happy to approve and merge

indica/readers/ppfreader.py
gadgil48: Yeah, I think this is a remnant of the old dictionaries where in ELEMENTS_BY_MASS, "calcium" and "cobalt" were commented out for what I suspect is this exact reason. Using Z-number and mass together should return a unique element, I'll add that in. Would the atomic number collection just be added to the same place where mass is recovered ~L250 and in the same fashion as well? I can't really test so I'll paste the line I want to add here and you can have a look to see if it's right, it would save me re-commiting and having to go through the CI again:
`atomic_num, atomic_num_path = self._get_signal(uid, instrument, "zqnn", revision)`
The new condition for checking would now be:
`if value[0] == atomic_num.data[0] and value[1] == int(round(mass.data[0]))`

elitherl: Firstly I'm being overkill suggesting "ZQNN" since I forgot about dtype "Z", which makes more sense when we just want Z...

Only thing that needs changing is to check against `int(atomic_num.data[0])` because for some reason it's returned as a float (don't know if that's a PPF or SAL issue though), otherwise I just tested it and it looks good

PR #71, opened 2021-09-22
Fixes dropped channel handling in utilities.input_check()

### LINKED ISSUE #66: Ignored channels incompatible with NaN checks in operators
Dropping channels when reading diagnostics substitutes values in the data with `nan` to show that it has been dropped, which then fails on `input_check` when setting up and operator (e.g. `FractionalAbundance`)

`nan` check is still required in the operators but the `dropped` attribute of data where channels have been dropped should be used to selectively allow intentional `nan`s

gadgil48:
Closes #66

### DIFF
--- a/indica/utilities.py
+++ b/indica/utilities.py
@@ -1,5 +1,6 @@
 """Various miscellanious helper functions."""
 
+from copy import deepcopy
 import inspect
 import string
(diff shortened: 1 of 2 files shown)


elitherl (approved):
Looks good to me

PR #74, opened 2021-09-29
Fixes mixing up of rows and columns for ADF15 data import

### LINKED ISSUE #73: ADAS reader mixes up rows and columns
**Describe the bug**
If nt == nd, reader often swaps dimensions. In other cases, data is mixed up and a simple transpose does not solve the issue.

**Minimal Reproducer**
Change of file reader function might solve the issue.

gadgil48:
Closes #73

### DIFF
--- a/indica/readers/adas.py
+++ b/indica/readers/adas.py
@@ -189,14 +189,11 @@ class ADASReader(BaseIO):
 
         """
 
-        def read_lines(f, nd, nt):
+        def explicit_reshape(data_to_reshape, nd, nt):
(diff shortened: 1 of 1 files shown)


gadgil48:
@marcosertoli , have a look to see if my modification is a suitable solution, I was more comfortable in making things as simple as possible so I let `np.fromfile()` do the importing work again and added the `explicit_reshape()` function which gives you more control over the decomposition of the 1d array into the 2d array than `np.ndarray.reshape()`

marcosertoli (approved):

PR #75, opened 2021-11-26
Add data source information to provenance

### LINKED ISSUE #52: Get latest PPF revision number
**Is your feature request related to a problem? Please describe.**
When asking for revision (sequence #) 0, the ppfreader doesn't return the effective revision, but keeps 0.

Read latest revision number instead of 0.

**Additional context**
call function in abstractreader (and then all readers) *_get_revision()*

elitherl:
Closes #52

Add diagnostic, uid, instrument, revision and quantitiy to
abstractreader provenance

In ppfreader get actual revision instead of potentially saving "0"

Make typing of revision in create_provenance not optional


elitherl:
I realised my test was trying to be too complicated, simplified it to something sensible and it's now passing so ready for review

marcosertoli (changes requested):
I don't agree create_provenance should be called in the specific readers e.g. ppfreader. It is a method of the abstractreader and it will fill in the provenance parameters that are read by the various readers. 
I would recommend adding a simple method to the ppt reader that updates the revision from 0 to the effective number. For ST40 and MDS+, I've added a method called _get_revision that returns the input if revision!=0, otherwise reads the effective revision from database and returns that. This is a few lines of code and leaves the provenance creation to the abstract reader

marcosertoli (approved):

marcosertoli:
All good from my side Ed. But I'd prefer you to deal with the conflicts on the fake_salclient. Once you've checked that I'll press the big green button

elitherl:
> All good from my side Ed. But I'd prefer you to deal with the conflicts on the fake_salclient. Once you've checked that I'll press the big green button

Thanks @marcosertoli, sure thing, I'll fix the conflicts and then we can merge

PR #80, opened 2021-12-21
Change labelling of start and end points for LoS transforms.

### LINKED ISSUE #64: Change LOS coordinates from (R,z,T) to (x,y,z)
**Describe the bug**
Current start and end LOS coordinates given in input the the abstract reader are in cartesian space, but the variable names are called (R,z,T) and thus misleading. The variable names should be changed to (x,y,z) where x=R, y=T, z=z to reflect the real meaning of the spatial coordinates.

**Expected behavior**
The functions/methods should work exactly as they do now, changing only the variable names and order in the calls to reflect the (x,y,z) convention.

**Additional context**
In the IDL code at the origin of Indica, only diagnostics viewing on a poloidal plane (R,z) were used. The need to expand the this further lead to the addition of the T coordinate, but the variable names were never changed.

gadgil48:
Closes #64

### DIFF
--- a/indica/converters/impact_parameter.py
+++ b/indica/converters/impact_parameter.py
@@ -68,7 +68,7 @@ class ImpactParameterCoordinates(CoordinateTransform):
             Tuple[DataArray, DataArray],
             lines_of_sight.convert_to_Rz(
                 coord_array(
-                    np.arange(len(lines_of_sight.R_start)), lines_of_sight.x1_name
+                    np.arange(len(lines_of_sight.x_start)), lines_of_sight.x1_name
                 ),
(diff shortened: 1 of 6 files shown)


marcosertoli (changes requested):
Haven't tried the functionality, but I'm sure it works fine. 
You've missed the re-labelling of a few internal variables though:

In convert_to_R, convert_from_Rz, distance: 
- R_s and R_e --> x_s and x_e
- T_s T_e -->  y_s_and y_e

Also, in the internal help of the class, there's still start and end of (R_, T_,  z_) instead of (x,y,z)

marcosertoli (approved):

PR #83, opened 2022-02-08
Extrapolate SXR impurity density and fit a perturbation

### LINKED ISSUE #23: Extrapolate impurity density beyond the detection limit of SXR
**Use bolometer to extend impurity density beyond SXR detection**
SXR detection limit at JET roughly coincides with an electron temperature of 1.5-2 keV. Beyond this limit, the SXR filter function is so steep that it cannot be trusted for absolute density measurements. For this reason, the high-Z impurity density calculated using equations 2.1 of [Journal of Plasma Physics 85 (2019) 905850504](https://doi.org/10.1017/S0022377819000618) must be truncated at a radius that coincides with this temperature limit.

**Proposed solution**
From the truncation position defined by the electron temperature limit, the user should have the choice to extrapolate in two manners (bool option in input). 

1. "simply" truncate the profile on the LFS and extend it using the shape of the electron density. 
2. from the truncation point extend the LFS mid plane values of the high-Z impurity density using a gaussian function with varying centroid position, amplitude and width

In order to calculate the 2D (R, z) map of the impurity density, one still needs the impurity's asymmetry parameter. This can be calculated in two ways (user choice, bool option in input):
a. using a measured toroidal rotation profile
b. calculating the toroidal rotation using the asymmetry parameter estimated from the truncated SXR 2D impurity density, and then extending this toroidal rotation profile to the separatrix using the electron density profile shape (same as in option 1).  

When the full 2D impurity density measurement has been computed, calculate the forward model of bolometer LOS-integrals using the resulting 2D high-Z impurity density and add the contributions from all other impurities. 

If option 2 has been chosen, the optimal parameters of the gaussian function can be estimated performing a least-squares fit comparing the forward model of the bolometer LOS to experimental values.

Imput parameters/keywords:
* truncation electron temperature threshold=2.e3 (eV): this is SXR filter dependent, so will change for different machines 
* toroidal rotation measurement=None: by default use self-consistent asymmetry calculated from SXR-estimated impurity density
* profile for extrapolation of asymmetry parameter=electron density: but it can be anything else if one has a alternative profile shape
* BOOL option to fit bolometer data viewing only LFS
* BOOL option to fit bolometer data normalised to most edge LFS in order to fit only the shape of the profile and not the absolute value

**Additional context**
Option 1 should be implemented even for cases where the poloidal asymmetry is extremely low. Option 2 is absolutely necessary for discharges with high toroidal rotation where most of the impurity density is trapped on the LFS midplane close to the pedestal top, often outside the detection limit of the SXR systems.

### LINKED ISSUE #95: rho vs rho_poloidal coordinate naming
There's a mismatch between use of `rho` and `rho_poloidal` for describing the same coordinate within the code, this includes in various checks and calculations when only one of the two will be compatible.

I think we either need to converge on a single naming convention for rho or implement more flexible checking/interpolating/calculations/etc... that can operate on rho or rho_poloidal depending on what's present (harder to do in my opinion). Thoughts?

Examples of `rho_poloidal`:
- `flux_surfaces.py`: `FluxSurfaceCoordinates` - Appends `kind` to "rho_" (e.g. "rho_poloidal" or "rho_toroidal") to make `x1` name, this name is then used to fetch coordinates from associated `DataArray` when doing coordinate transforms
- `utilities.py`: `coord_array` - Sets datatype name according to coord_name provided, "norm_flux_pol" datatype name is only for `rho_poloidal`
- `spline_fit.py`: `SplineFit` - First argument is expected to be `("norm_flux_pol", "plasma")` (as set by `coord_array`), but isn't if `rho` is used instead of `rho_poloidal`

Examples of `rho`:
- `impurity_concentration.py`: `ImpurityConcentration` - Assumes input parameters have dimension `rho`, fails if not present (due to `rho_poloidal` being used instead)
- `extrapolate_impurity_density.py`: `ExtrapolateImpurityDensity` - Same as `ImpurityConcentration`

gadgil48:
Closes #23 
(Should also partially close #95 )


elitherl (changes requested):
Added notes from discussion yesterday, only a few things need changing and then it looks good

There are a couple of places where `loc` is used to select from coordinates which can fail if the user supplied coordinate arrays don't contain an exact value, I've noted some but will add more if I find any I've missed

elitherl (approved):

PR #84, opened 2022-02-21
Finished x,y,z labelling conversion for abstractreader and ppfreader

### LINKED ISSUE #82: Complete xyz conversion
**Describe the bug**

**Proposed solution**

**Additional features**

gadgil48:
Partly closes #82


elitherl (approved):
Changes all look good, and can confirm this fixes the issue blocking SXR reading.
Just need to tweak PR settings so that issue #82 isn't closed in the process

PR #96, opened 2022-03-18
Jmwood94 0/82 line of sight transform

### LINKED ISSUE #82: Complete xyz conversion

jmwood94-0:
Closes #82 

Change in numpy_typing.py needed --> lines_of_sight.py imports LabelledArray. Since numpy.typing import ArrayLike works (line 17), then np not declared and import fails. 

All other changes are made to line_of_sight.py and test script.


marcosertoli:
Passing all pre-commit, fixed various parts of the code and changed a few variable names.

Only test that is not passing (see below), please ask @gadgil48 :
tests/unit/converters/test_line_of_sight.py:7: error: Cannot find implementation or library stub for module named 'tests.unit.test_equilibrium_single'

Last thing on the TODO list: 
- write a few tests for the single functions of the class 
- test a few cases where it might fail
- demonstrate extreme cases do what you expect, e.g. parallel lines don't have an intersection

jmwood94-0:
@marcosertoli 

I've added the tests that you recommended above. I bypassed the equilibrium import issue by copying the code directly from "test_equilibrium_single" - we can replace this if there is a way we can import the function from the other script.

jmwood94-0:
@gadgil48 @elitherl 

Hi Both, I wonder if you can help. Marco and I are ready to do our final checks before finishing the merge pull request but there's something weird going on with the automatic checks. Earlier the checks went through correctly [75724a4] but are still waiting for the checks to run after the latest changes to the project.

Is it possible we can run these check?

PR #98, opened 2022-03-18
Upgrade NumPy and Pillow to resolve security alerts from Dependabot

### LINKED ISSUE #97: Resolve Dependabot security alerts for NumPy and Pillow
With Pillow version: 8.4.0 and NumPy version: 1.19.5
(Low, Moderate and Critical are severity of security concerns according to Dependabot)
`Dependabot` alerts:
- Infinite loop in Pillow: Low
- Arbitrary expression injection in Pillow: Critical
- Path traversal in Pillow: Moderate
- Out-of-bounds Read in Pillow: Moderate
- Improper Initialization in Pillow: Critical
- Buffer Overflow in NumPy: Moderate

Suggested fix: Update `pyproject.toml` and run `poetry lock` again.

gadgil48:
Closes #97

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,10 +40,10 @@ readme = 'README.md'  # Markdown files are supported
 include = [".git", "indica/git_version"]
 
 [tool.poetry.dependencies]
-python = "^3.6.1"
+python = "^3.8"
 prov = "^1.5.3"
 netCDF4 = "^1.5.4"
-numpy = "^1.19.1"
+numpy = "^1.21.0"
 scipy = "^1.5.2"
 xarray = "0.16.0"
 matplotlib = "^3.3.0"
--- a/indica/numpy_typing.py
+++ b/indica/numpy_typing.py
@@ -5,16 +5,15 @@ labeled dimensions and to scalar.
 
 """
 
+import numbers
 import typing
 
+import numpy as np
 import xarray
 
 try:
     from numpy.typing import ArrayLike
 except ImportError:
-    import numpy as np
-    import numbers
-
     ArrayLike = typing.Union[
         numbers.Number, typing.Sequence[numbers.Number], np.ndarray
     ]
--- a/poetry.lock
+++ b/poetry.lock
@@ -127,22 +127,6 @@ optional = false
 python-versions = ">=3.6"
 
 [[package]]
-name = "dataclasses"
-version = "0.8"
-description = "A backport of the dataclasses module for Python 3.6"
-category = "main"
-optional = false
-python-versions = ">=3.6, <3.7"
-
-[[package]]
-name = "decorator"
-version = "4.4.2"
-description = "Decorators for Humans"
-category = "main"
-optional = false
-python-versions = ">=2.6, !=3.0.*, !=3.1.*"
-
-[[package]]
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -40,10 +40,10 @@ readme = 'README.md'  # Markdown files are supported
 include = [".git", "indica/git_version"]
 
 [tool.poetry.dependencies]
-python = "^3.6.1"
+python = "^3.8"
```
k-collie, line 43: Could we cope with ^3.7? I'm not sure Heimdall has 3.8.

.pre-commit-config.yaml
```diff
@@ -25,6 +25,6 @@ repos:
       - flake8-import-order
       name: "Check Code Style Using flake8"
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812  # Use the sha / tag you want to point at
+    rev: v0.782  # Use the sha / tag you want to point at
```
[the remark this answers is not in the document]
gadgil48: To bring it in line with the version number in `pyproject.toml`. Running the newer version locally causes it throw errors because it says it doesn't recognise any typing for `scipy` modules.

indica/converters/line_of_sight.py
```diff
@@ -116,16 +116,17 @@ def __eq__(self, other: object) -> bool:
...
+        result = result and np.all(self.x_start == other.x_start)  # type: ignore
+        result = result and np.all(self.z_start == other.z_start)  # type: ignore
+        result = result and np.all(self.y_start == other.y_start)  # type: ignore
+        result = result and np.all(self.x_end == other.x_end)  # type: ignore
+        result = result and np.all(self.z_end == other.z_end)  # type: ignore
+        result = result and np.all(self.y_end == other.y_end)  # type: ignore
+        result = result and np.all(self.dl == other.dl)  # type: ignore
+        result = result and np.all(self.x2 == other.x2)  # type: ignore
+        result = result and np.all(self.R == other.R)  # type: ignore
+        result = result and np.all(self.phi == other.phi)  # type: ignore
```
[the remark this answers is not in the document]
gadgil48: Yep I agree

pyproject.toml
gadgil48: Unfortunately not, there are too many features and dependencies in the code that rely on python 3.8. Even though the requirement in the file was 3.6.1 before, all CI tests were run on python 3.8. An example of the features seems to be unpacking operations inside a tuple ie. a return statement that returns a tuple which has a unpack operation for one of its elements - see line 499 in `indica/operators/invert_radiation.py` but there are other bits as well (like specific types from `typing`). I think @elitherl is running InDiCA on Heimdall so I guess he could guide you through setting up a version python 3.8.

elitherl: Yep, I'm running on heimdall with a python 3.9.10 install. I've found [pyenv](https://github.com/pyenv/pyenv) to be the most convenient way to maintain user installs of different python version, they also have an [automatic installer](https://github.com/pyenv/pyenv-installer) that handles most things and talks you through the rest of the set up, but feel free to drop me a message if you run into any issues.

k-collie (approved):

PR #109, opened 2022-04-01
Jmwood94 0/82 line of sight transform

### LINKED ISSUE #82: Complete xyz conversion

jmwood94-0:
Closes #82 

Minor changes to convert methods and x1, x2 labels. 

Commits passed local pre-commit checks.

### DIFF
--- a/indica/converters/line_of_sight.py
+++ b/indica/converters/line_of_sight.py
@@ -96,8 +96,8 @@ class LinesOfSightTransform(CoordinateTransform):
         self.x_end = DataArray(x_end)
         self.z_end = DataArray(z_end)
         self.y_end = DataArray(y_end)
-        self.x1_name = name + "_coords"
-        self.x2_name = name + "_los_position"
+        self.x1_name = "channel"
+        self.x2_name = "los_position"
 
         # Set "dl"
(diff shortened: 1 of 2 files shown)


marcosertoli (changes requested):
Minor change Jon from this:
        _, self.z = self.convert_to_Rz(0, x2, 0)
         self.R = np.sqrt(self.x**2 + self.y**2)

to this:
        self.R, self.z = self.convert_to_Rz(0, x2, 0)

No need to recalculate it outside of the function!

marcosertoli (approved):

jmwood94-0:
@elitherl @marcosertoli 

I have pushed updates as requested, and unfortunately failed the tests - but do not see how I could have done this. The error occurs inside test_ppf_reader? I wonder if either of you could help me! Otherwise the tests look to be successful. 

There is also a fault with codecov/project which I do not understand.

marcosertoli (approved):

marcosertoli:
All fine. The coverage had diminished, but that was to be expected.

I've merged the changes.

Have a good weekend,
Marco

PR #111, opened 2022-04-01
Marcosertoli/110 abstractreader

marcosertoli:
Modified accordingly as explained in the issue. Dependencies should have been taken care of, no tests written yet, but since it's the readers we're talking about, we can leave them for the time being.


marcosertoli:
- Tests: I still don't understand the way Chris wrote the tests, so cannot modify them. All that's missing to get the tests running is that the dictionary returned by the ppfreader includes a "revision" element (see example given by st40reader in the branch marcosertoli/st40). While someone on the UKAEA side fixes this, I can fix the codecov/patch (with separate test scripts, not modifying Chris' original...).
- codecov/project: coverage should be tied to moving the datatypes to a separate function

marcosertoli:
All tests passing apart from read/write from/to cache for ppf_reader. Everything updated to current master apart from intended changes! Ready for merging.

indica/readers/abstractreader.py
```diff
@@ -41,9 +43,9 @@ class DataReader(BaseIO):
 
     This defines the interface used by all concrete objects which read
     data from the disc, a database, etc. It is a `context manager
-    <https://docs.python.org/3/library/stdtypes.html#typecontextmanager>`_
+    <https://protect-eu.mimecast.com/s/f7vJCpzxoFzjOpcPXjtX?domain=docs.python.org>`_
```
elitherl, line 46: These URLs should be changed back, not sure where they lead but I don't have access to it so I suspect others won't either

indica/readers/abstractreader.py
```diff
@@ -544,20 +519,33 @@ def get_equilibrium(
...
+            print(f"No data from {uid}.{instrument}:{revision}")
+            return database_results
+        _revision = database_results["revision"]
+
         diagnostic_coord = "rho_poloidal"
         times = database_results["times"]
-        downsample_ratio = int(
-            np.ceil((len(times) - 1) / (times[-1] - times[0]) / self._max_freq)
-        )
+        times_unique, ind_unique = np.unique(times, return_index=True)
```
elitherl, line 548: Should probably keep the `downsample_ratio` feature and add the unique time check as well, rather than instead of

indica/readers/abstractreader.py
```diff
@@ -1168,30 +1174,702 @@ def _get_bremsstrahlung_spectroscopy(
...
+            Vertical location of stop positions for lines of sight for this data.
 
         """
         raise NotImplementedError(
             "{} does not implement a '_get_spectroscopy' "
             "method.".format(self.__class__.__name__)
         )
 
+    #
+    # def get_astra(
```
elitherl, line 1703: Other parts of `get_astra` are being added in this PR, can this be uncommented and included too?

indica/readers/abstractreader.py
```diff
@@ -544,20 +519,33 @@ def get_equilibrium(
...
+        global_quantities = {
+            "psi",
+            "rmag",
+            "zmag",
+            "rgeo",
+            "faxs",
+            "fbnd",
+            "ipla",
+            "wp",
+            "df",
```
elitherl, line 531: This extra quantity ("df") is the reason the `PPFReader` tests are failing; `df` doesn't exist for EFIT so isn't available in the test data set.
What is `df`? Is it ST40 specific? If so, can it be moved into `_IMPLEMENTATION_QUANTITIES` for the specific reader and that will also fix the failing tests

elitherl (commented):
New `abstractreader` methods look good, and thanks for the various bits fixed throughout!

One general comment: `test_abstract_reader_pytest.py` is intended to replace the original `test_abstract_reader.py`, I think it would make sense to keep the same name to make that clear, what do you think?

Just a couple of comments on specific bits for things that could be changed (or at least discussed quickly) - including the failing `PPFReader` tests - then should be good to merge

indica/readers/abstractreader.py
elitherl: Add comment describing state of `get_astra` (works but no tests)

marcosertoli (commented):
All good. Modified what needed to be modified and will push the changes in a moment.

Since you have deleted the original test, I've changed the name of the new taking away the _pytest.

Also, one thing that's still missing is the conversion of the readers to give position and direction of the lines of sight so we can apply Jon's new los-transform across the board. This should be a relatively easy Job that Kingsley could do and would require:
- change the outputs of the ppfreader so that the dictionary has "position" and "direction" of the lines of sight and not the start/end.
- in the abstract reader swap the old for the new los-transform
-  swap from DataArray + attributes including error and transform to a DataSet with value, error, transform - this applies to all diagnostics, including those with multiple lines-of-sight

It's worth tackling all of this in one go. And since this should not affect the end user, it can be done ASAP after the beta release. But should be done with a bit of urgency before anything else is modified.

elitherl (approved):

PR #117, opened 2022-04-07
Install instructions

k-collie:
Will need to be updated to point to point to the tag on release as part of the release process but this works until a first release.


k-collie:
Thinking about it more it would probably be best to suggest using poetry to install (so we use poetry.lock), we can be more flexible once we have tested on different versions.

k-collie:
Switched to poetry install, removed link checking from the CI (since it can be affected by the websites we link to) and restricted indica to python 3.8 or 3.9 until #127.

PR #122, opened 2022-04-08
Skip no-commit-to-master check in CI.

k-collie:
Should fix linting failing when merging PRs into master after #120 was merged.


k-collie:
Seems to work: https://github.com/ukaea/Indica/runs/5889162649?check_suite_focus=true

gadgil48 (approved):

PR #127, opened 2022-04-26
Update minimum versions of scipy and mypy

### LINKED ISSUE #126: Installing dependencies fails
Tested on fresh Ubuntu 22.04 install, python 3.10.4, pip 22.0.4, setuptools 62.1.0.

`scipy==1.5.3` and `typed-ast==1.4.3` fail to build, there are no wheels available.

**Reproduce:**

`pip install scipy==1.5.3`

**Possible Solutions**

- Increase minimum versions of these packages.
- Add necessary packages/libraries (python-dev, LAPACK/BLAS) to build dependencies to install guide

k-collie:
Fixes #126 

Bump scipy and mypy versions that were causing problems, move and fix mypy config.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,11 +40,11 @@ readme = 'README.md'  # Markdown files are supported
 include = [".git", "indica/git_version"]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.11"
 prov = "^1.5.3"
 netCDF4 = "^1.5.4"
 numpy = "^1.21.0"
-scipy = "^1.5.2"
+scipy = "^1.6.2"
 xarray = "0.16.0"
 matplotlib = "^3.3.0"
 sal = "^1.2.2"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -40,11 +40,11 @@ readme = 'README.md'  # Markdown files are supported
 include = [".git", "indica/git_version"]
 
 [tool.poetry.dependencies]
-python = "^3.8"
-prov = "^1.5.3"
+python = ">=3.8,<3.11"
```
k-collie, line 43: `scipy` started adding maximum python versions in 1.6.2, so `poetry` now complains if we use the `^` here

k-collie: This is an unnecessary restriction of poetry that I don't think `pip` has, every dependency version must be compatible with every python version.

pyproject.toml
```diff
@@ -40,11 +40,11 @@ readme = 'README.md'  # Markdown files are supported
 include = [".git", "indica/git_version"]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.11"
 prov = "^1.5.3"
 netCDF4 = "^1.5.4"
 numpy = "^1.21.0"
-scipy = "^1.5.2"
+scipy = "^1.6.2"
```
k-collie, line 47: 1.6.2 is the first release which specifies a maximum python version

elitherl (approved):
Update looks good

PR #151, opened 2022-06-27
Accept rho, theta in extrapolate_impurity_density.

### LINKED ISSUE #150: Allow extrapolate impurity density taking rho, theta rather than R, z
`ExtrapolateImpurityDensity` currently only takes `impurity_density_sxr` on R, z. Would be useful if it could operate taking rho, theta.

k-collie:
Fixes #150 

Needs more work to verify asymmetry_from_rho_theta.
Currently returning centre full on NaNs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ indica/git_version
 *.png
 *.prof
 *.lprof
+.ropeproject/
--- a/indica/operators/atomic_data.py
+++ b/indica/operators/atomic_data.py
@@ -144,10 +144,10 @@ class FractionalAbundance(Operator):
 
         inputted_data = {}
 
-        input_check("Ne", Ne, DataArray, greater_than_or_equal_zero=True)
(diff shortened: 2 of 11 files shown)


k-collie:
`extrapolator.asymmetry_modifier` has nans in same places as returned density.

.gitignore
```diff
@@ -16,3 +16,4 @@ indica/git_version
 *.png
 *.prof
 *.lprof
+.ropeproject/
```
gadgil48, line 19: Was this a temporary change or is it still necessary?

k-collie: They're files generated by a refactoring tool which shouldn't be committed afaict so I don't think it's necessary but doesn't hurt either.

gadgil48 (approved):
Looks good, if possible some more tests to satisfy `codecov` would be nice (just for the sake of green tick OCD) but if it's not possible/not useful then it's fine (we can always reset `codecov` before beta release anyway)

PR #154, opened 2022-07-04
Add a warning when some quantities less than 0

### LINKED ISSUE #153: Add warnings in reader for invalid data
Many quantities should be >= 0 for example. Warn when invalid values are detected.

k-collie:
Fixes #153.

### DIFF
--- a/indica/readers/abstractreader.py
+++ b/indica/readers/abstractreader.py
@@ -13,6 +13,7 @@ from typing import Iterable
 from typing import List
 from typing import Set
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


elitherl:
Looks good, it's good to have that warning. Up to you if it's a separate PR or part of this one, but I think it'd be worth adding missing `>0` checks to operators (e.g. `optimize_perturbation` since that's how we discovered this), which should be a stopping check rather than a warning

elitherl (approved):

PR #170, opened 2022-08-15
Add mean_charge argument to AsymmetryParameter

### LINKED ISSUE #168: Add support for calculated mean charge in asymmetry parameter
`AsymmetryParameter` uses mean impurity charge in calculation, but currently takes a value from stored elemental data. This assumes element is fully ionised and a uniform mean charge across plasma, however a more accurate mean charge can be calculated and should be included to improve accuracy of `AsymmetryParameter`

elitherl:
Adds mean_charge to `__call__` method, expects DataArray with coords "element", "rho_poloidal", and "time" (however only "element" is strictly necessary). Replaces use of fixed mean_charge.

Fixes #168


k-collie:
Can we add this fix to `ToroidalRotation.__call__` too? I think it just does the reverse of `AsymmetryParameter` but has the same problem.

k-collie (approved):

k-collie:
Input checks need updating for  #151

indica/operators/centrifugal_asymmetry.py
```diff
@@ -135,20 +149,21 @@ def __call__(  # type: ignore
 
         # mypy on the github CI suggests that * is an Unsupported operand type
         # between float and DataArray, don't know how to fix yet so for now ignored
-        toroidal_rotation = 2.0 * ion_temperature * asymmetry_parameter  # type: ignore
-        toroidal_rotation /= impurity_mass * (
-            1.0
-            - (mean_charge * main_ion_mass * Zeff * electron_temp)  # type: ignore
-            / (impurity_mass * (ion_temperature + Zeff * electron_temp))
+        return (
```
gadgil48, line 152: The change to using `np.sqrt` and `spc.speed_of_light` is good, but it might be better for readability to have the calculation done in stages on a single variable (`toroidal_rotation`) as before. Is there any specific reason to do the calculation in the `return` statement?

indica/operators/centrifugal_asymmetry.py
```diff
@@ -244,33 +262,38 @@ def __call__(  # type: ignore
...
 
         ion_temperature = ion_temperature.sel(element=impurity)
 
-        c = 3.0e8  # speed of light in m/s
-        toroidal_rotations /= c
-
         # mypy on the github CI suggests that * is in an Unsupported operand type
         # between float and DataArray, don't know how to fix yet so for now ignored
-        asymmetry_parameter = (
+        return (
```
gadgil48, line 291: Same comment as earlier in the code, is there any reason to do the calculation in the `return` statement?

tests/unit/operators/test_centrifugal_asymmetry.py
```diff
@@ -365,6 +383,14 @@ def test_centrifugal_asymmetry():
         dims=["element", "rho_poloidal", "t"],
     )
 
+    mean_charges = DataArray(
+        data=np.swapaxes(
+            np.tile(np.array([5, 4, 3, 2, 1]), (len(elements), len(t), 1)), 1, 2
+        ),
+        coords=[("element", elements), ("rho_poloidal", rho_profile), ("t", t)],
+        dims=["element", "rho_poloidal", "t"],
+    )
```
k-collie, line 392: Tile array and multiply by a factor for each element (tungsten should have higher mean charge than beryllium).

elitherl: Could use `ELEMENTS` (from `datatypes.py`) to choose suitable charge values for a given element - it's max charge rather than mean charge but that's wouldn't make a difference for the tests

indica/operators/centrifugal_asymmetry.py
k-collie: I found it easier to compare to equation 2.12 written like this. I also didn't like that we kept in-place modifying the `toroidal_rotation` variable which only had the correct value when it was returned.

indica/operators/centrifugal_asymmetry.py
k-collie: Here the in-place modify method led to us mutating `toroidal_rotations` which is an input value and so might have side-effects.

PR #171, opened 2022-08-15
Add separate gaussian fitting to high and low field sides

### LINKED ISSUE #169: Add split high/low field side Gaussian perturbation
Current implementation of a Gaussian perturbation (in `optimize_perturbation`) fits a single Gaussian on the low field side and uses the impurity asymmetry to map this to the high field side. As the Gaussian perturbation is confined to the region outside SXR validity, if asymmetry from SXR is used this value will be unreliable. Suggest fitting the high field side Gaussian with an independent amplitude to allow its asymmetry to be independently derived from the bolometry fit.

elitherl:
Fit 2 gaussian features with independent amplitudes, sharing a common position and width. High field side gaussian shares upper bound with low field side and has a 0 amplitude lower bound

Fixes #169

### DIFF
--- a/indica/operators/extrapolate_impurity_density.py
+++ b/indica/operators/extrapolate_impurity_density.py
@@ -791,18 +791,31 @@ class ExtrapolateImpurityDensity(Operator):
                 for the selected time point.
                 xarray.DataArray with dimensions (channels)
             """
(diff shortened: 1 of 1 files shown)


elitherl:
See #174 for improved solution to the same problem

PR #174, opened 2022-08-24
Fit asymmetry parameter in optimize_perturbation

### LINKED ISSUE #169: Add split high/low field side Gaussian perturbation

k-collie:
Fixes #169. Fits the asymmetry parameter in the region specified as invalid.

### DIFF
--- a/indica/operators/extrapolate_impurity_density.py
+++ b/indica/operators/extrapolate_impurity_density.py
@@ -947,7 +947,7 @@ class ExtrapolateImpurityDensity(Operator):
                     lower_amp_bound,
                     lower_width_bound,
                     lower_pos_bound,
-                    0.1 * asymmetry_parameter.min(),
+                    -10 * np.abs(asymmetry_parameter.max()),
                 ]
             ),
             np.array(
@@ -955,7 +955,7 @@ class ExtrapolateImpurityDensity(Operator):
                     upper_amp_bound,
                     upper_width_bound,
                     upper_pos_bound,
-                    10 * asymmetry_parameter.max(),
+                    10 * np.abs(asymmetry_parameter.max()),
                 ]
             ),
         ]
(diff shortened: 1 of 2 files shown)


gadgil48 (approved):
All good, ready to merge

indica/operators/extrapolate_impurity_density.py
```diff
@@ -891,17 +947,29 @@ def objective_func(objective_array: Sequence, time: float):
...
                     lower_pos_bound,
+                    0.1 * asymmetry_parameter.min(),
                 ]
             ),
             np.array(
                 [
                     upper_amp_bound,
                     upper_width_bound,
                     upper_pos_bound,
+                    10 * asymmetry_parameter.max(),
```
k-collie, line 958: @elitherl Do these bounds seem good? Maybe we should subtract something from the lower bound in case it is small and positive to allow it to be negative?

elitherl: Could base the bounds on some factor (e.g. 5, 10) * the absolute value of the maximum, to account for negative asymmetry

k-collie: This seems like a good idea but I am now struggling to get the optimizer to converge for isel(t=1).

k-collie: Got it to converge by tweaking `xtol` up a bit, hopefully the changes to `x_scale` should make `xtol` better behaved.

gadgil48 (approved):

PR #176, opened 2022-08-25
Update SXR LOS calculation

### LINKED ISSUE #175: Correct and simplify SXR LOS calculation
Change SXR LOS calculation to match SURF, simplifies calculation in process

elitherl:
Fixes #175 

Change calculation of LOS angle to match SURF calculation
Simplify calculation steps from original


k-collie:
I did this on pen and paper to confirm, this calculation is correct.

k-collie (approved):

PR #178, opened 2022-09-15
Expand CXRS reading

elitherl:
Adds additional CXRS DDAs and concentration reading capabilities

### DIFF
--- a/indica/readers/ppfreader.py
+++ b/indica/readers/ppfreader.py
@@ -120,6 +120,30 @@ class PPFReader(DataReader):
             "zefh": ("effective_charge", "plasma"),
             "zefv": ("effective_charge", "plasma"),
         },
+        **{
+            "cx{}m".format(val): {
+                "angf": ("angular_freq", "ions"),
+                "ti": ("temperature", "ions"),
+                "conc": ("concentration", "ions"),
+            }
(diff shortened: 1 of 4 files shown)


k-collie:
I think we could add a few things to `test_get_charge_exchange` and `ppf_samples.json`.

indica/readers/ppfreader.py
```diff
@@ -94,11 +94,22 @@ class PPFReader(DataReader):
...
+            for val in ("s", "d", "f", "g", "h")
+        },
+        **{
+            "cx{}6".format(val): "get_charge_exchange"
+            for val in ("s", "d", "f", "g", "h")
+        },
+        **{
+            "cx{}4".format(val): "get_charge_exchange"
+            for val in ("s", "d", "f", "g", "h")
+        },
```
k-collie, line 112: code golf option:

```suggestion
        **{"cx{}".format(val) for val in product(("s", "d", "f", "g", "h"), ("4", "6", "m"))}
```

elitherl:
@k-collie As part of this change I've also made `conc` a PPF specific dtype following discussion yesterday. The current implementation can be be trimmed down a bit if we make the changes to how it uses default/global quantities

k-collie (approved):

PR #180, opened 2022-09-16
Use poetry 1.1.15 in CI - fix broken install

### LINKED ISSUE #179: Poetry broken in CI
See https://github.com/python-poetry/poetry/issues/6300.

Solution: Update poetry.

k-collie:
Fixes #179 

Tested with poetry 1.2.0 but could not install indica.


elitherl (approved):
I read somewhere that poetry 1.2 changes the how `.lock` files work so suspect that's why it wouldn't work, 1.1.15 sounds like a sensible version to go to

PR #182, opened 2022-09-20
Add workflow classes

elitherl:
Adds classes to handle higher level operations for impurity analysis - combines use of existing operators and maintains a self-consistent view of the plasma being described.

Currently contains some changes in other pull requests, and probably contains changes that could be factored out into their own, stand alone pull requests (TBD).


elitherl:
Also currently includes workaround for #183 - needs fixing before merge

indica/workflow/observers.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class Observable:
+    """
+    Class to hold some data, with information about Observer classes, and
+    signal observers to update when stored data has changed
+    """
+
+    def __init__(self, initial_value: Optional[Any] = None):
+        self.observers: List[TypeObserver] = []
```
k-collie, line 48: I think instead of `TypeObserver` and `TypeObservable` we can put the types in quotes like this:

```suggestion
        self.observers: List['Observer'] = []
```

k-collie (changes requested):
Here's some preliminary comments on the JET and base workflow. We could think about moving some things into the base workflow like:

- authentication (setup steps)
- get_diagnostics
- initialise default values
- average powerloss

indica/workflow/base_workflow.py
```diff
@@ -0,0 +1,581 @@
...
+        self.n_high_z = self._rescale_n_high_z()
+        return self.n_high_z
+
+    def _optimise_n_high_z(self) -> DataArray:
+        raise NotImplementedError(
+            f"{self.__class__} does not implement _optimise_n_high_z method"
+        )
+
+    def optimise_n_high_z(self) -> DataArray:
+        self.n_high_z = self._optimise_n_high_z()
```
k-collie, line 499: As much as it pains me we should probably stick to [oxford spelling](https://en.wikipedia.org/wiki/Oxford_spelling), or at least be consistent.

k-collie:
We should make how the data is binned/moving-averaged an option too.

indica/workflow/jet_workflow.py
```diff
@@ -0,0 +1,699 @@
...
+            # Calculate initial high-z element density and extrapolate
+            print(template.format("calculate_n_high_z"))
+            self.calculate_n_high_z()
+            print(template.format("extrapolate_n_high_z"))
+            self.extrapolate_n_high_z()
+            derived_emissivity = self.electron_density * (
+                self.ion_densities * self.sxr_power_loss_charge_averaged
+            ).sum("element")
+            if np.any(derived_emissivity > self.sxr_emissivity):
+                raise UserWarning(
```
k-collie, line 113: we need to `warnings.warn("message", WarningType)` rather than `raise` warnings

PR #187, opened 2022-09-20
Line of sight with conversion to rho

marcosertoli:
Added assignment of flux surface transform object to perform conversion to rho.

Included functionality to mapping a desider profile to the LOS and integrate it.

For completeness, both lines_of_.. and line_of_sight have been modified --> we should aim to convert all our workflows to line_of_sight and delete lines_of_sight from the project.

### DIFF
--- a/indica/converters/__init__.py
+++ b/indica/converters/__init__.py
@@ -7,8 +7,8 @@ from .enclosed_volume import EnclosedVolumeCoordinates
 from .flux_major_radius import FluxMajorRadCoordinates
 from .flux_surfaces import FluxSurfaceCoordinates
 from .impact_parameter import ImpactParameterCoordinates
-from .lines_of_sight import LinesOfSightTransform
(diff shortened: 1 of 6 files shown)


marcosertoli:
The code is ready to be reviewed, but is still missing new tests for the new methods of the line_of_sight transform as well as its implementation across the diagnostics that are currently only available for JET (e.g. ECE, TS, CXRS, etc.).

Remains to be defined how to implement the line_of_sight (singular) for those diagnostics that have multiple lines of sight.

Only error that I wasn't able to clarify in the pre-commit is:
*indica/readers/abstractreader.py:504:5: C901 'DataReader.get_equilibrium' is too complex (20)*

k-collie:
JET has VJAC but AJAC but PSIN is labelled "PSNI" so we can account for that in `IMPLEMENTATION_QUANTITIES` in `ppfreader`.

PR #198, opened 2022-10-28
Handle 0 densities in asymmetry_from

### LINKED ISSUE #197: asymmetry_from cannot handle 0 density
The resulting asymmetry will be `nan` if the low field side asymmetry passed to `asymmetry_from_rho_theta` or `asymmetry_from_R_z` is 0. This could be handled by making the asymmetry parameter 1 at these coordinates.

k-collie:
Fixes #197


k-collie:
Should it be 0 instead, implying no asymmetry?

PR #206, opened 2022-12-09
Add more appropriate argument types for spline fit

### LINKED ISSUE #183: SplineFit argument types too restrictive
`SplineFit` operator currently accepts arguments as below:

```python
ARGUMENT_TYPES: List[Union[DataType, EllipsisType]] = [                                                                                                               
    ("norm_flux_pol", "plasma"),                                                                                                                                      
    ("time", "plasma"),                                                                                                                                               
    ("temperature", "electrons"),                                                                                                                                     
    ...,                                                                                                                                                              
]
```

which restricts use to electron temperature only - `SplineFit` is usable for any quantity with coordinates of `norm_flux_pol` and `time`. Argument type checking needs updating to be more generic

k-collie:
Fixes #183

### DIFF
--- a/indica/operators/spline_fit.py
+++ b/indica/operators/spline_fit.py
@@ -20,13 +20,11 @@ from xarray import DataArray
 
 from .abstractoperator import EllipsisType
 from .abstractoperator import Operator
-from .abstractoperator import OperatorError
 from .. import session
 from ..converters import bin_to_time_labels
 from ..converters import CoordinateTransform
 from ..converters import FluxSurfaceCoordinates
 from ..datatypes import DataType
-from ..datatypes import GeneralDataType
 from ..numpy_typing import ArrayLike
 from ..utilities import broadcast_spline
 from ..utilities import coord_array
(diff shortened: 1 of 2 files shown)


k-collie:
Remove default value from bounds and make this a parameter in `SplineFit` that gets passed through.

indica/operators/spline_fit.py
```diff
@@ -153,8 +154,35 @@ def __init__(
             raise ValueError(
                 "lower_bound must be either a scalar or array of same size as knots"
             )
+        self.bc_type = bc_type
         self.spline: Spline
         self.spline_vals: DataArray
+
+        accepted_general_datatypes = [
```
elitherl, line 161: Suggestion from discussion: could replace with keys from datatypes dictionary

elitherl (changes requested):
Suggestion from discussion: modify `ARGUMENT_TYPES` in `__call__` method to fetch current type from first instance of `data` given

PR #208, opened 2022-12-12
Allow manually triggered CI runs

### LINKED ISSUE #207: Allow manually triggering CI runs
This is useful if the job fails spontaneously.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

k-collie:
Fixes #207

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,6 +3,7 @@
 name: sphinx docs
 
 on:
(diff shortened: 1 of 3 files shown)


elitherl (approved):
Makes sense to add

PR #210, opened 2023-01-06
St40

marcosertoli:
Major modifications to:
- geometry operators, including the abstractoperator, line_of_sight, transect, etc.
- atomic_data, with the capability to interpolate rates instead of calculating values every time
- simplified equilibrium (why would we need an abstract class for this?), volume and area calculation
- reader and available_quantities to get single diagnostics and not sets (e.g. for SXR available_quantities should be the brightness from one camera, not a camera name)
- diagnostic and plasma models, including data from O. Marchuck on He-like Ar X-ray spectra

Failing some tests and missing quite a few for the new modules, but on track to develop or finalize them.


marcosertoli:
Closing draft pull request. Will submit separate pull requests for separate parts of the code

PR #217, opened 2023-02-09
Corrected the rbnd and zbnd calculation

### LINKED ISSUE #213: Boundary RBND and ZBND in equilibrium_dat method of test_equilibrium_single
**Describe the bug**
RBND and ZBND are not correct, spanning only 90 degrees instead of 360. 

If this fake equilibrium is used to generate an Equilibrium object and flux_coords is called, all rho values below the plasma x-point are put to zero. The parameter that is checked for this is the z position of the x-point, which will now extend to any (R,z) below the z position of the outer midplane.

**Expected behavior**
RBND and ZBND should go a full "circle" for the method to work correctly

gadgil48:
Closes #213 

Modifies the `rbnd` and `zbnd` calculations so that the values for the full extent of theta are recovered.
Example at the end of `indica/models/equilibrium.py` is purely for diagnostic purposes feel free to remove.

Most if not all of the changes to `indica/models/plasma.py` are due to pre-commit pipeline.

### DIFF
--- a/indica/models/equilibrium.py
+++ b/indica/models/equilibrium.py
@@ -1,11 +1,11 @@
-
+import matplotlib.pyplot as plt
 import numpy as np
 import xarray as xr
 
 from indica.converters import FluxSurfaceCoordinates
 from indica.converters import TrivialTransform
-from indica.equilibrium import Equilibrium
 from indica.converters.time import get_tlabels_dt
+from indica.equilibrium import Equilibrium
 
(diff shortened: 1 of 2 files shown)


marcosertoli:
I've checked out only the equilibrium.py file and merged into st40 branch to avoid merging in also unwanted mods to plasma.py.

@gadgil48 , you can delete this branch if you wish

PR #219, opened 2023-02-21
Marcosertoli/186 los transform rho

marcosertoli:
Modifications to Equilibrium and LOS transform and all the methods and classes that depend on it.

Cascade to JET-specific methods and classes still to be performed, as fixing of dedicated tests.


indica/converters/abstractconverter.py
```diff
@@ -310,3 +371,168 @@ def encode(self) -> str:
...
+        ``((xmin, xmax), (zmin, zmax)``. Defaults to values for JET.
+
+    Returns
+    -------
+    start_coordinates
+        A Tuple (1x3) giving the X, Y and Z start positions of the line-of-sight
+    end_coordinates
+        A Tuple (1x3) giving the X, Y and Z end positions of the line-of-sight
+    """
+
```
gadgil48, line 463: Both the `line` and `extremes` functions seem unnecessary as they don't really reduce the number of lines or simplify a complicated piece of code that's needed multiple times. If it's a question of masking the underlying `np.linspace` and `np.array` functions with readable names like `line` and `extremes` then perhaps it's okay though.

indica/converters/abstractconverter.py
```diff
@@ -310,3 +371,168 @@ def encode(self) -> str:
...
+            boundaries = {
+                "x_in": x_plasma_inner,
+                "x_out": x_plasma_outer,
+                "y_in": y_plasma_inner,
+                "y_out": y_plasma_outer,
+            }
+        return boundaries, angles, rho_equil
+
+
+def find_wall_intersections(
```
gadgil48, line 429: Any reason for the `origin` and `direction` arguments to be type `Tuple`? `LabeledArray` or `ArrayLike` might be better.

indica/utilities.py
```diff
@@ -20,6 +20,7 @@
 from xarray.core.variable import Variable
```
gadgil48, line 20: Most of the added functions here seem useful (maybe with the exception of `print_like` and `get_function_name`) but it would be helpful to include docstrings that are consistent with the `numpy` style for all of the added functions. Also unit tests for some of the more complicated functions like `intersection`.

indica/models/equilibrium.py
```diff
@@ -0,0 +1,289 @@
+import numpy as np
```
gadgil48, line 1: Agreed that having a model of a fake equilibrium is definitely useful, but the naming of this file is a bit confusing. Even though it's in a different location I have confused this file with `indica/equilibrium.py` multiple times, perhaps `example_equilibrium.py` or something similar would be better. Not particularly important, just a suggestion.
Also docstrings for the functions are needed, although if these functions are from the fake equilibrium that I wrote then feel free to create an issue for adding the docstrings and I'll make the changes and open a pull request to merge into `Marcosertoli/186 los transform rho`.

indica/converters/line_of_sight.py
```diff
@@ -4,18 +4,22 @@
 from typing import cast
```
gadgil48, line 4: A few functions need docstrings but otherwise looks good!

tests/unit/converters/test_line_of_sight.py
```diff
@@ -2,193 +2,136 @@
```
gadgil48, line 2: Some functions are missing from these unit tests: `convert_from_Rz`, `convert_to_rho`(though this is already noted in the code), `map_to_los`, `check_rho_and_profile`, `integrate_on_los` and `calc_impact_parameter`. I don't know if any unit test for `plot_los` or `example_run` is necessary but I leave that up to you.

tests/unit/converters/test_line_of_sight.py
```diff
@@ -2,193 +2,136 @@
...
-        assert np.isclose(dl, dl_out, rtol=1e-2)
+    assert np.abs(dl - los.dl) < 1.0e-4
 
-        if debug:
-            print(f"x2 = {x2}")
-            print(f"dl_out = {dl_out}")
+    if debug:
+        print(f"x2 = {los.x2}")
+        print(f"dl_out = {los.dl}")
     return
```
gadgil48, line 137: The unit test `test_intersections` can probably be moved to `tests/unit/test_utilities.py` now.

gadgil48 (commented):
Besides what I've mentioned in the comments I've included, the rest looks good!

marcosertoli:
Taking current version of new los transform with 2D capabilities and testing it on st40 branch.

Will make a new pull request for this soon.

PR #228, opened 2023-03-28
Michaelgemmell/magnetic recon diagnostic

michael-gemmell:
I added a diagnostic object for comparing EFIT quantities like wp.

### DIFF
--- a/indica/models/magnetic_recon.py
+++ b/indica/models/magnetic_recon.py
@@ -1,0 +1,87 @@
+import xarray as xr
+import matplotlib.pyplot as plt
+
+from indica.models.abstractdiagnostic import DiagnosticModel
+from indica.models.plasma import example_run as example_plasma
+from indica.readers.available_quantities import AVAILABLE_QUANTITIES
+
+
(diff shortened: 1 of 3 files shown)


indica/models/magnetic_recon.py
```diff
@@ -0,0 +1,87 @@
+import xarray as xr
+import matplotlib.pyplot as plt
+
+from indica.models.abstractdiagnostic import DiagnosticModel
+from indica.models.plasma import example_run as example_plasma
+from indica.readers.available_quantities import AVAILABLE_QUANTITIES
+
+
+class MagneticRecon(DiagnosticModel):
```
marcosertoli, line 9: I would change the name to **EquilibriumReconstruction**.
We can envisage it to calculate all the relevant equilibrium quantities including, in the far future, the internal equilibrium reconstruction (hence the name) using the EFIT boundary, similar to ASTRA or TRANSP.

marcosertoli (approved):
Apart from the name-change, necessary for future expansion, looks good and I'd merge it in as soon as you're ready. We can expand the scope in the coming months.

PR #230, opened 2023-04-04
Marcosertoli/st40 delete unused tests

marcosertoli:
Cleaning up of most tests and modules to make pre-commit and all tests work on the st40 branch:
- Deleting all modules and tests that are not used in the st40 branch
- Simplification of complex modules e.g. abstractreader, atomic_data
- Fixing all pre-commit issues
- Fixing and simplifying all relevant tests


marcosertoli:
All tests passing locally, but need to check with Sanket or Kingsley to solve the lower level checks from CI/CD (poetry upgrade, coverage report, sphinx).

Once that's sorted, we could merge in st40 with all CI/CD working.

marcosertoli:
Deleted unnecessary CI actions of documentation and upload of code coverage to codecov. Both are to be reinstated when merging into master.

indica/models/helike_spectroscopy.py
michael-gemmell, line 1: What is the rationale behind using interp here instead of sel? Would we ever want to model on a different time vector then plasma.t?

michael-gemmell (approved):

PR #235, opened 2023-05-03
Michaelgemmell/bayes dev

michael-gemmell:
Adding logic to the bayes class such that channels are considered individual measurements. Also includes some unit tests for the Plasma class.

### DIFF
--- a/indica/models/helike_spectroscopy.py
+++ b/indica/models/helike_spectroscopy.py
@@ -281,7 +281,10 @@ class Helike_spectroscopy(DiagnosticModel):
                 calc_rho=calc_rho,
             )
             # LOS integral removes NaNs so add them back
-            self.measured_spectra = self.measured_spectra.where(self.measured_spectra != 0, np.nan)
+            self.measured_spectra = self.measured_spectra.where(
+                self.measured_spectra != 0, np.nan
+            )
+
     def _calculate_temperatures(self):
         x1 = self.los_transform.x1
         x1_name = self.los_transform.x1_name
@@ -763,6 +766,7 @@ def helike_LOS_example(nchannels=3):
         passes=1,
     )
     return los_transform
+
 
 def example_run(pulse: int = 9229, plasma=None, plot=False, **kwargs):
     # TODO: LOS sometimes crossing bad EFIT reconstruction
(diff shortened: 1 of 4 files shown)


indica/bayesmodels.py
```diff
@@ -73,17 +83,25 @@ def _build_bckc(self, params: dict, **kwargs):
     def _ln_likelihood(self):
         ln_likelihood = 0
         for key in self.quant_to_optimise:
-            # TODO: What to use as error?  Assume percentage error if none given...
-            # Float128 is used since rounding of small numbers causes
-            # problems when initial results are bad fits
-            model_data = self.bckc[key].values.astype("float128")
+            # Float128 since rounding of small numbers causes problems
```
marcosertoli, line 86: Question:
would rounding the model data be a better/cleaner solution instead of changing type?

indica/bayesmodels.py
```diff
@@ -73,17 +83,25 @@ def _build_bckc(self, params: dict, **kwargs):
...
             exp_data = (
-                self.data[key]
-                .sel(t=self.plasma.time_to_calculate)
-                .values.astype("float128")
+                self.data[key].sel(t=self.plasma.time_to_calculate).astype("float128")
             )
-            _ln_likelihood = np.log(gaussian(model_data, exp_data, exp_data * 0.10))
-            ln_likelihood += np.nanmean(_ln_likelihood)
+
+            exp_error = exp_data * 0.10  # Assume percentage error if none given.
```
marcosertoli, line 93: Suggestion:
I would move the percentage_error value (0.1 presently) to the class init and make it a class attribute. This way you can decide.

indica/models/helike_spectroscopy.py
```diff
@@ -280,6 +280,10 @@ def _calculate_los_integral(self, calc_rho=False):
                 t=self.spectra.t,
                 calc_rho=calc_rho,
             )
+            # LOS integral removes NaNs so add them back
```
marcosertoli, line 283: Comment/suggestion:
I don't get this, and have tried it out and it seems this .where() is not required. 
Could it be legacy and could it be deleted?

marcosertoli (changes requested):
Minor comments, suggestions and questions only.

I don't see any clashes with the mods I'm proposing on the other pull request, but you might want to address this pull before looking into mine.

PR #262, opened 2023-07-13
Michaelgemmell/bayes workflow

michael-gemmell:
Main inclusion here is the AbstractBayesWorkflow class and two examples under bayes_worklow_dev and bayes_workflow_example. These include all the writing to MDS+ scripts.
Still missing tests and fixing old tests.


marcosertoli (commented):
First comments

jmwood94-0 (approved):

jmwood94-0 (approved):

PR #264, opened 2023-07-18
Marcosertoli/satakihino

marcosertoli:
I’ve fixed some bugs in the workflow and it now works ok, but neither Aleksandra nor I don’t know how to plot Zeff, phantom and inferred result, with the confidence interval as you do for the other graphs. 

Could you modify this branch so it plots this? Then please send it back to me so by reviewing I can understand what you’ve done and do further mods myself :-)!

Thanks,
Marco


michael-gemmell (commented):
I added the zeff to the profiles that are saved during the optimisation.

For some reason I can't see where the plotting function is but in there you have to add zeff as a 1d_plot.

PR #268, opened 2023-07-25
Marcosertoli/reorganize bremss calculation

marcosertoli:
There might be some merge conflicts, will check and modify accordingly, then you can merge it in.


marcosertoli:
All good to go

PR #271, opened 2023-07-27
Jmwood94 0/107 add neutral beam

jmwood94-0:
Pull request for adding a new Neutral Beam model class to Indica and working on forward modelling using the existing Charge Exchange class on ST40 data. 

The main changes include: 

- New Neutral Beam class which stores useful input data, no call functions yet implemented. 
- Significant update to the ChargeExchange class, adding call to FIDASIM inside the __call__ function and a new method that calls and runs a local version of FIDASIM. 
- An example workflow for running the FIDASIM forward model using ST40 data. 

There are some places highlighted in the code where code hasn't yet been implemented, or I have had to use a work around to bypass some python environment issues I have been facing.


marcosertoli:
Added most recent changes to st40 in a separate pull request that @michael-gemmell is going to look into, as it changes some of his previous work.
Once that is taken care of, I'll tackle this pull request.

PR #272, opened 2023-07-28
Marcosertoli/zeff profile

marcosertoli:
Adding:
- new reading for pre-analysis spectroscopy MDS+ data
- preliminary @satakihino work on Zeff calculation from data from PI passive spectroscopy using inversion & inference

@michael-gemmell , please do look at this pull request if you have time already today or Monday, so you can better help @satakihino in solving the last issues on Bayes workflow for Zeff calculation using experimental data instead of phantoms


marcosertoli:
Ready to merge, no more changes planned!

indica/models/diode_filters.py
```diff
@@ -63,6 +64,70 @@ def __init__(
...
+        )
+
+        # Take away neutron spikes in pixel intensity
+        _spectra = spectra.sortby("wavelength").sel(wavelength=wavelength_slice)
+        _spectra = _spectra.where(
+            xr.ufuncs.fabs(_spectra.diff("wavelength", n=2)) < 0.4e-3
+        )
+
+        # Fit spectra to calculate background emission, filter and integrate
+        if fit_background:
```
michael-gemmell, line 117: This should probably be it's own function

indica/models/plasma.py
```diff
@@ -470,20 +460,33 @@ def initialize_variables(self, tstart: float, tend: float, dt: float):
         )
 
     def assign_profiles(
-        self, profile: str = "electron_density", t: float = None, element: str = "ar"
```
michael-gemmell, line 473: Can remove element from the function call

michael-gemmell (approved):

indica/models/diode_filters.py
marcosertoli: Absolutely! Where would you put it? Shall we make a new file in indica/operators/ for spectroscopy methods? 

I'll add a TODO

indica/models/plasma.py
marcosertoli: I wouldn't do that because this way we retain the possibility to assign different shapes to the different impurities. But I just saw an error which I will fix once I've merged in and send it back to you for confirmation.

PR #274, opened 2023-08-02
Marcosertoli/astra plotting

marcosertoli:
Added standalone preliminary file with methods to plot results from Indica workflows. 

Currently plots time evolution of several quantities & HDA results. Still to be refined, but at least provides reference for future development.


marcosertoli:
Methods in this pull request are stand-alone and preliminary. They don't affect any other method or workflow.

Merging in so it can be used and/or checked by others + fed back in the discussion to decide how to best implement this.

PR #275, opened 2023-08-04
Marcosertoli/read spd

marcosertoli:
New reader for LINES and SPD using new MDS+ structures


marcosertoli:
Linting and tests are all that's needed. 
I still have to solve the docs building, but it's not functional so can wait.

Once you've reviewed, you can either send it back to me if you have comments and/or suggestions, or squash and merge directly.

PR #286, opened 2023-09-14
BLOM_XY1 reader

marcosertoli:
Added BLOM_XY1 reader. TODO: channel mapping still unclear and untouched. SXRC channel mapping also internally reorganised and to be sorted in MDS+.


marcosertoli:
Merging it in to provide further consistency checks vs. AXUV diodes. @ccolgan , please do review and propose changes when you have time.

PR #293, opened 2023-10-12
Michaelgemmell/st40 dev

michael-gemmell:
Complete refactor of the bayesian workflow to use dependency injection i.e. external objects handle the methods and these objects can be initialised / changed separately from the workflow.

For example instead of :
workflow.setup_optimiser()
workflow.optimiser_do_thing()
->
optimiser_object = ExampleOptimiser()
worfklow.optimiser = optimiser_object
workflow.optimiser.do_thing()

This PR is currently unfinished and I'm actively fixing / developing it.


michael-gemmell:
The main body of code to look at is under indica/workflows/bayes_workflow_example

PR #294, opened 2023-10-26
Bda

michael-gemmell:
**Working** version of BDA - let me know if you have feedback or questions.


indica/bayesblackbox.py
```diff
@@ -0,0 +1,143 @@
+from copy import deepcopy
+import warnings
+from flatdict import FlatDict
+import numpy as np
+from scipy.stats import uniform
+np.seterr(all="ignore")
+warnings.simplefilter("ignore", category=FutureWarning)
+
+def gaussian(x, mean, sigma):
```
marcosertoli, line 9: we have too many of gaussian() functions across the project...
profiles_gauss has a quite general version of it, including background, amplitude, width and centroid, but it's missing the normalisation that you have in this version...
Can we generalise it and put in just 1 place, e.g. utilities or physics?

indica/bayesblackbox.py
marcosertoli, line 1: I'll leave the rest of the review to Jussi for now!

indica/workflows/abstract_bayes_workflow.py
```diff
@@ -150,330 +46,270 @@ def _build_result_dict(self):
...
-            .sel(rho_poloidal=0, method="nearest")
-            .std(dim="index"),
-            "NI0_ERR": self.blobs["ion_density"]
-            .sel(element=self.plasma.main_ion)
-            .sel(rho_poloidal=0, method="nearest")
-            .std(dim="index"),
+                .sel(element=self.plasma_context.plasma_settings.main_ion)
+                .sel(rho_poloidal=0, method="nearest")
+                .median(dim="index"),
+            "WP": self.blobs["wp"]
```
marcosertoli, line 243: why median and not mean?

indica/workflows/abstract_bayes_workflow.py
marcosertoli, line 1: as for black box part, I'll look into it again after Jussi's review!

indica/workflows/bayes_plots.py
marcosertoli, line 1: Is there a way we can abstract the plotting? 
I've been already experimenting with refactiring to have long_name and units embedded in the DataArrays which makes plotting labelling way easier. Worth sitting down and sharing views on this so we can simplify and compress the plotting functions as much as possible.

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+
+DEFAULT_DIAG_NAMES = [
+    "xrcs",
+    "efit",
+    "smmh1",
+    "cxff_pi",
+    "ts",
+]
+
+FAST_DIAG_NAMES = [
```
marcosertoli, line 135: what are all of these FAST_ global variables for? they're not used anywhere. Worth deleting them completely?

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+        print("Data strategy: Phantom data")
+        return self.model_callable()
+
+    def process_data(self, model_callable: Callable):
+        self.pre_process_data(model_callable)
+        self.opt_data = flatdict.FlatDict(self.binned_data, ".")
+
+
+@dataclass
+class MockData(PhantomData):
```
marcosertoli, line 691: What's the difference between Mock and Phantom? Phantom is taking experimental equilibrium but mock data? Mock is all Mocked??
Also, if we need to abstract (I hear you on this!), is there more we can abstract?

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+        if missing_transforms:
+            raise ValueError(f"Missing transforms: {missing_transforms}")
+
+        self.transforms = self.diagnostic_transforms
+        self.binned_data: dict = {}
+        self.raw_data: dict = {}
+
+
+@dataclass
+class OptimiserEmceeSettings:
```
marcosertoli, line 715: Why do we need a dataclass for this? 
In this application, what does having a class give us that a dictionary doesn't?

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+    sample_method: str = "random"
+    starting_samples: int = 100
+    stopping_criteria: str = "mode"
+    stopping_criteria_factor: float = 0.01
+    stopping_criteria_sample: int = 20
+    stopping_criteria_debug: bool = False
+
+
+@dataclass
+class OptimiserContext(ABC):
```
marcosertoli, line 730: Is this aimed at providing a Context generalized to different types of optimisers and you're currently applying it to the Emcee, building the desired methods for the use of that package?

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+        max_iter = self.optimiser_settings.iterations * self.optimiser_settings.nwalkers
+        npad = ((0, int(max_iter * (1 - self.optimiser_settings.burn_frac) - post_sample.shape[0])), (0, 0))
+        results["post_sample"] = np.pad(post_sample, npad, constant_values=np.nan)
+
+        results["auto_corr"] = self.autocorr
+        return results
+
+
+@dataclass
+class BayesBBSettings:
```
marcosertoli, line 868: You'll have to explain this to me, please!

indica/workflows/bayes_workflow.py
```diff
@@ -0,0 +1,1065 @@
...
+    @abstractmethod
+    def data_strategy(self):
+        return None
+
+    def _check_if_data_present(self, data_strategy: Callable = lambda: None):
+        if not self.binned_data:
+            print("Data not given: using data strategy")
+            self.binned_data = data_strategy()
+
+    def pre_process_data(self, model_callable: Callable):
```
marcosertoli, line 612: I don't get this and will need some more explanation, please!

indica/writers/bda_tree.py
marcosertoli, line 1: Absolutely fine with this for the time being, but I think we need to review this again and then go to DVM!

marcosertoli (approved):
Looks great Michael. A lot of work and really interesting solutions!

I'll try it out on specific pulses and example in the next weeks so can then give you some hands-on driven feedback before Jussi comes back with his structural review.

marcosertoli:
I've also fixed the read_st40 conflict, by merging the fix for missing revisions that was implemented in the st40 branch.
Should now work across the board, but please check!

indica/workflows/abstract_bayes_workflow.py
michael-gemmell: If you have a skewed distribution the mean might be misleading. I feel it is more intuitive to think of Bayesian inference in terms of quantiles i.e. median, 25-75% etc. I'm curious if you have a reason for preferring the mean?

PR #302, opened 2023-12-08
Marcosertoli/spline fit rshift

marcosertoli:
Not elegant, and not fast, but it spits out some useful results. I've tested it on several pulses including 11089, 11314, 11316, ...

See if it's at all useful.


michael-gemmell:
image
I think there must be some difference between our envs if this is working for you but not me.

marcosertoli:
My current env is built following the Readme on the ST40 branch.
Worth upgrading to that if you're still on an older environment.
All of your packages for BDA were included. If not, it's worth adding packages via poetry so toml and lock files are updated and everyone can upgrade their environment!

PR #307, opened 2024-01-24
Any reader import fails without MDSplus package

elitherl:
Importing any of the reader classes currently fails if you don't have the MDSplus package installed due to importing ST40Reader. MDSplus isn't readily available (e.g. on PyPI) so we shouldn't expect users to have it installed if they're not using a reader that requires it

### DIFF
--- a/indica/models/plasma.py
+++ b/indica/models/plasma.py
@@ -19,7 +19,6 @@ from indica.operators.atomic_data import PowerLoss
 import indica.physics as ph
 from indica.profiles_gauss import Profiles
 from indica.readers import ADASReader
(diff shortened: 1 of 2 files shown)


marcosertoli:
OK with intermediate solution.
One of the tasks we have to do in the next few months is to modularise further the readers, including extracting a SALReader from the PPFReader, and the MDSReader from the ST40Reader, and making all tests using MockReaders.

Let's discuss the details on Friday.

marcosertoli (approved):

elitherl:
Modularising readers makes sense to me, makes sense as a proper solution

PR #314, opened 2024-03-29
Marcosertoli/development

marcosertoli:
Changed AVAILABLE_QUANTITIES and DATATYPES to make it easier to understand what is required of the reader, to assign long_name and units to each quantity read, and to simplify the reading.

Abstract reader now stripped to the bones, but still requiring work:
- ASTRA (it's a mess)
- EFIT (could be simplified further)

Once st40reader has been refactored to extract all MDS functions, it can also be refactored to make it much clearer what is required when building a reader.


marcosertoli:
Following changes to DATATYPES and AVAILABLE_QUANTITIES, changed Indica-native DataArray formatting and adapted Plasma class accordingly.

michael-gemmell (commented):
Looks reasonable to me!

PR #315, opened 2024-04-04
Jussi reader separation

hakosaj:
Pre-commit fails due to a very minor linting thing (how the config files are structured causes this).

Removed the duplicate and ran pre-commit, some other minor structure fixes too.

There is a server error: not sure if it is my environment or the structure is still off. Can you test this?

I think all strictly-MDS+ methods are now in the MDSutils. At least thats the case as far as I understand? YOu said there was still something, but I do not know what it is. All the methods that directly access MDS+ are in the utils, and the methods in readed call these utils methods. That is what the plan was no?


marcosertoli:
Fixed minor bugs in the mds-utils, solved linting. Ready to merge from my perspective, but eager to have @michael-gemmell  have a look and a go with anything that's not currently in the testing suite.

marcosertoli:
Organisational comment: should we save all the CONF files in a separate directory, e.g. indica/configurations/?

marcosertoli (approved):

PR #317, opened 2024-04-24
Bda merge

michael-gemmell:
- Plasma Ti no longer has element as a dimension
- Updates to BDA tree structure
- BDA running using config files
- Tests now include bda_run (working on CI/CD!)


indica/workflows/bayes_workflow.py
```diff
@@ -441,38 +472,55 @@ def fit_ts_profile(self, data_context, split="LFS", quant="ne"):
...
         fit = xr.where(fit < 0, 1e-10, fit)
-        return fit
+        return fit, processed_data
 
-    def set_ts_profiles(self, data_context, split="LFS"):
+    def set_ts_profiles(self, data_context, split="LFS", R_shift=0.0):
 
-        ne_fit = self.fit_ts_profile(data_context, quant="ne", split=split) * 1e19
-        te_fit = self.fit_ts_profile(data_context, quant="te", split=split) * 1e3
+        ne_fit, ne_data = self.fit_ts_profile(
```
marcosertoli, line 479: It's probably worth adding a PPTS reader so that you can use the PPTS fits saved to MDS+. It's anyway the same fitting routine, and it's more consistent.

tests/unit/models/test_plasma.py
```diff
@@ -1,82 +0,0 @@
-import numpy as np
-
```
marcosertoli, line 2: Was this the only test on the Plasma class? If so, might it be best to reinstate some test for this class later on (not in this pull request)?

marcosertoli (approved):
Good to go from my end!

PR #318, opened 2024-04-24
Marcosertoli/phantom reader

marcosertoli:
Name doesn't fully represent the pull request.

Details of the pull request:
- A draft (standalone) phantom reader is now there, but doesn't affect BDA or other workflows.
- Finalize DATATYPES and UNITS.
- Delete unused objects and methods, including Provenance that has been totally deleted from the project.
- Delete unused tests.
- New defaults/ directory with pkl files of default objects (Equilibrium, Transforms, Plasma) to make use of tests/development on phantoms faster + methods to generate them and load them.


indica/models/bolometer_camera.py
```diff
@@ -186,7 +185,17 @@ def example_run(
 ):
 
     if plasma is None:
-        plasma = example_plasma(pulse=pulse, calc_power_loss=True)
+        from indica.equilibrium import fake_equilibrium
```
michael-gemmell, line 188: imports at the tops

michael-gemmell (commented):
Can you explain the reason for these default objects?

marcosertoli:
Reasoning is that every time we want to do any Phantom analysis or tests on CI/CD we can load a default Plasma, Equilibrium and a set of Transforms for all diagnostics without having to generate them on the fly. That saves time during tests, as well as development, especially on the Plasma side, and gives us an equilibrium that is more relevant than the current fake_equilibrium that is not fit for purpose.

indica/models/plasma.py
```diff
@@ -1191,20 +835,121 @@ def __call__(self):
...
     impurities: Tuple[str, ...] = ("c", "ar", "he"),
-    impurity_concentration: Tuple[float, ...] = (0.03, 0.001, 0.01),
-    verbose: bool = True,
-    n_rad: int = 41,
-    full_run=False,
-    calc_power_loss: bool = False,
+    load_from_pkl: bool = True,
     **kwargs,
 ):
+    from pathlib import Path
```
michael-gemmell, line 937: .

marcosertoli:
I had thought of this as being the interface between plasma and profilers. If you decouple them, would you need another class to couple them again?

P.S. ...I keep on having issues with circular imports and would really appreciate if you had some insights on good practice to avoid them from the start!

indica/readers/modelreader.py
michael-gemmell, line 1: I think this is good but this reader doesn't need to be machine or model specific. It can take a dictionary of {name: diagnostic_model} in the initialisation. 
The way it is done here is more like a normal reader where diagnostics are hard coded - but this should be less like a reader and more like a model co-ordinator.

michael-gemmell:
> Reasoning is that every time we want to do any Phantom analysis or tests on CI/CD we can load a default Plasma, Equilibrium and a set of Transforms for all diagnostics without having to generate them on the fly. That saves time during tests, as well as development, especially on the Plasma side, and gives us an equilibrium that is more relevant than the current fake_equilibrium that is not fit for purpose.

We have to make sure that a new instance of these are created every test so that we are testing properly then.

michael-gemmell:
> I had thought of this as being the interface between plasma and profilers. If you decouple them, would you need another class to couple them again? P.S. ...I keep on having issues with circular imports and would really appreciate if you had some insights on good practice to avoid them from the start!

We would need another class, interface or method to assign the values to the plasma.
This is an example of where it is good to separate the logic from the pipeline in my opinion. looking like {profile_name: profiler} -> ProfileManager -> profiles. It is the same pattern as the other context objects (we should discuss this).

I'd need to see where the circular imports are happening to give my opinion, but general advice is circular dependencies are a sign of over-coupling.

indica/readers/modelreader.py
marcosertoli: Absolutely! I've currently hardwired the models, but a default set can be in the Conf, and otherwise I agree it can go in the init.

marcosertoli:
I've addressed the comments and fixed another minor detail. 

@michael-gemmell, feel free to push the big green button and merge it in if you approve.

PR #319, opened 2024-05-01
Updated default ADF11 files to ones with highest Te range and added f…

marcosertoli:
…unction to tabulate cooling factors.


indica/operators/atomic_data.py
```diff
@@ -916,3 +920,125 @@ def default_profiles(n_rad: int = 100):
...
+    new_figure: bool = True,
+    include_neutrals: bool = False,
+):
+    """
+    Initialises atomic data classes with default ADAS files and runs the
+    __call__ with default plasma parameters
+    """
+    tau = None
+    Ne_const = 5.0e19
+    Nh1 = 1.0e17
```
michael-gemmell, line 938: Can these be kwargs instead of hard coded?

michael-gemmell (approved):

PR #320, opened 2024-05-10
Refactored documentation deleting wrong/old docs and leaving only the…

marcosertoli:
Refactored documentation deleting wrong/old docs and leaving only the API.


pixelifytica:
I'm a bit hesitant to delete those `.gitkeep` files because I think they have something to do with building the docs, otherwise I agree with the trimming down. There's stuff in there that I think will be worth resurrecting further down the line (like the input data section being adapted into a description of what the readers can return, for example)

marcosertoli:
@pixelifytica , I'll work on this pull request again once we've transitioned to 311

pixelifytica (approved):

pixelifytica:
Changes included in PR #338

PR #321, opened 2024-05-20
Mg/ppts reader

michael-gemmell:
Added PPTS reader methods to abstractreader and st40reader.

Some modifications to how the filtering is done was necessary as it was binning empty time slices.
Now filtering happens before the binning and is saved in self.filtered_data


indica/readers/abstractreader.py
```diff
@@ -185,6 +185,54 @@ def _get_thomson_scattering(
...
+            elif "_rho" in quantity:
+                dims = ["t", "rho_poloidal"]
+            elif "_data" in quantity:
+                dims = ["t", "channel"]
+            elif "rshift" in quantity:
+                dims = ["t"]
+            else:
+                raise ValueError(f"Unknown quantity: {quantity}")
+
+            data[quantity] = self.assign_dataarray(
```
marcosertoli, line 212: For the data profiles (ne_data, te_data, ...) you should add the additional dimensions rho_data, R_data and z_data to channel.

You're reading them in the st40_reader, but not assigning them here :-).

marcosertoli (changes requested):
Looks good! I'd just add the additional dimensions to the _data profiles and then it's good to go from my end

indica/readers/abstractreader.py
michael-gemmell: I've just added them as attributes, not sure if there is a better way to handle it with xarray.

marcosertoli:
You should use assign_coords e.g. in the ADAS adf15 reader:

        pecs = pecs.assign_coords(wavelength=("index", wavelength))

This way you can use swap_coords method to change in-between coordinates.

For your case, you'll have "channel" instead of "index",  and rho_poloidal instead of wavelength. And you'll have to call this three yimes so you can also add R and Z as additional coordinates.

Marco


michael-gemmell:
Thanks! I always immediately forget this after I'm done.

PR #322, opened 2024-05-21
r_shift/z_shift are now DataArrays with time coord

michael-gemmell:
To allow time dependent r shifts I made R_shift either a DataArray with a time coordinate or a float. This DataArray is then interpolated onto the equilibrium time vector.

This seems to work but please give it a look as I'm not that familiar with the Equilibrium class.


michael-gemmell:
Currently broken something in the transforms...

michael-gemmell:
Commented out test_calculate_zeff as it doesn't work with new R_offset

indica/equilibrium.py
```diff
@@ -34,24 +35,22 @@ class Equilibrium:
...
         :py:meth:`~indica.readers.DataReader.get_equilibrium`. TODO: List full set
         of required quantities.
-    R_shift : float
+    R_shift
         How much to shift the equilibrium inwards (or the remapped diagnostic outwards)
-        on the major radius.
-        TODO: this and z_shift should be time-dependent...
-    z_shift : float
+        on the major radius. Either a float for all time slices or a DataArray
+        with coord time
```
marcosertoli, line 41: worth specifying coord name is "t" so it's clear what the input should be?

indica/equilibrium.py
```diff
@@ -70,6 +69,18 @@ def __init__(
         )
         self.psi = equilibrium_data["psi"]
 
+        if isinstance(R_shift, float):
+            R_offset = xr.DataArray(np.full_like(self.t, R_shift), coords=(self.t,))
```
marcosertoli, line 73: I'd recommend using this:
`xr.full_like(self.t, R_shift)`

indica/equilibrium.py
```diff
@@ -533,21 +542,26 @@ def flux_coords(
...
+        _R = R + self.R_offset
+        _z = z + self.z_offset
+
         if t is not None:
             check_time_present(t, self.t)
             rho = self.rho.interp(t=t, method="nearest")
             R_ax = self.rmag.interp(t=t, method="nearest")
             z_ax = self.zmag.interp(t=t, method="nearest")
             z_x_point = self.zx.interp(t=t, method="nearest")
+            _z = _z.interp(t=t, kwargs={"fill_value": 0})
```
marcosertoli, line 554: you shouldn't need the "fill_value" since the interpolation should not work if the time requested is not present in self.t, so you can just call it as `.interp(t=t)`

indica/equilibrium.py
```diff
@@ -533,21 +542,26 @@ def flux_coords(
...
+            _R = _R.interp(t=t, kwargs={"fill_value": 0})
+
         else:
             rho = self.rho
             R_ax = self.rmag
             z_ax = self.zmag
             t = self.rho.coords["t"]
             z_x_point = self.zx
+            _z = _z.interp(t=t, kwargs={"fill_value": 0})
```
marcosertoli, line 563: same as previous comment

indica/equilibrium.py
```diff
@@ -34,24 +35,22 @@ class Equilibrium:
...
+        with coord time
 
     """
 
     def __init__(
         self,
         equilibrium_data: Dict[str, DataArray],
-        R_shift: float = 0.0,
-        z_shift: float = 0.0,
+        R_shift: Union[DataArray, float] = 0.0,
```
marcosertoli, line 52: I'd make a new datatype in indica.numpy_typing (call it something like FloatDataarray?) so you can import it and there's no need to write Union[..] in the other calls either.

indica/operators/gpr_fit.py
```diff
@@ -243,12 +243,19 @@ def gpr_fit_ts(
...
     dt = 0.01
     st40 = ReadST40(pulse, tstart, tend, dt)
-    st40(instruments=["ts", "efit"], R_shift=0.00)
+    st40.get_equilibrium()
+    st40(
+        instruments=[
+            "ts",
+            "efit",
+        ],
+        R_shift=xr.DataArray(data=0.02, coords=(st40.equilibrium.t,)),
```
marcosertoli, line 257: this as above `xr.full_like(st40.equilibrium.t, 0.02)`

indica/readers/read_st40.py
```diff
@@ -1,4 +1,5 @@
 from copy import deepcopy
+from typing import Union
```
marcosertoli, line 2: not needed if you make a new numpy_typing type

indica/readers/read_st40.py
```diff
@@ -110,8 +111,8 @@ def get_equilibrium(
         self,
         instrument: str = "efit",
         revision: RevisionLike = 0,
-        R_shift: float = 0.0,
-        z_shift: float = 0.0,
+        R_shift: Union[DataArray, float] = 0.0,
```
marcosertoli, line 114: as above

indica/readers/read_st40.py
```diff
@@ -316,7 +317,7 @@ def __call__(
         tstart: float = None,
         tend: float = None,
         dt: float = None,
-        R_shift: float = 0.0,
+        R_shift: Union[DataArray, float] = 0.0,
```
marcosertoli, line 320: as above

tests/integration/workflows/test_calculate_zeff.py
marcosertoli, line 1: why doesn't it work? if you roll out this and the Zeff workflow crashes, it's a bit of a mess :-o...

marcosertoli (changes requested):
minor changes, but I need the Zeff workflow to work, otherwise I can't hope to implement it in PPAC anytime soon 👍 !

tests/integration/workflows/test_calculate_zeff.py
michael-gemmell: You are doing things with the equil.R_offset assuming it's a float. That's why the code breaks.

marcosertoli (approved):

PR #324, opened 2024-05-23
Marcosertoli/ppts inconsistencies

marcosertoli:
Minor mods + (temporary) fix to Zeff workflow. 

Still can't get BDA test to run correctly...

@michael-gemmell: can you please check and modify accordingly?


michael-gemmell:
> Minor mods + (temporary) fix to Zeff workflow.
> 
> Still can't get BDA test to run correctly...
> 
> @michael-gemmell: can you please check and modify accordingly?

The issue is how to bin when using a single timepoint. Before I guess it was at minimum using 2 time points.

marcosertoli (commented):
Looks like a good fix to me.

If you agree with it, you can press the big green button.

PR #326, opened 2024-06-04
Equilibrium shifting R / z in initialisation

michael-gemmell:
Simple fix to move the equilibrium psi and other quantities based on a user supplied R_shift / z_shift


michael-gemmell:
Whatever is going wrong with the tests needs to be debugged. 

What do you think should be filled here? or should it be kept as NaNs
image

marcosertoli (approved):

PR #309, opened 2024-06-12
PPFReader methods need an update

pixelifytica:
As in title, `PPFReader` methods are out of sync with how `DataReader` works. I've fixed a few but will work though them, mostly in order that I use them, and fix those to merge back in.

Methods:
  * [x] `_get_charge_exchange`
  * [x] `_get_thomson_scattering`
  * [x] `_get_equilibrium`
  * [x] `_get_cyclotron_emissions`
  * [x] `_get_radiation`
  * [x] `_get_bremsstrahlung_spectroscopy`


pixelifytica:
`PPFReader` methods should now all be up to date and working properly, needed to add the `idlbridge` as an optional dependency due to how the JET CXRS system stores geometry information - not only installed when using the "JET" extras group, never by default.

Still need to double check geometry being read is working correctly before merging but thought I'd put the PR up now so you can check over in the mean time @marcosertoli

pixelifytica:
Also closes #313

indica/operators/spline_fit.py
```diff
@@ -107,6 +106,8 @@ def example_run(
     knots: list = None,
     plot: bool = True,
 ):
+    from indica.readers.read_st40 import ReadST40
```
marcosertoli, line 109: For info, I'm moving all the examples to a new /examples/ directory, so we can decouple the drivers from the examples. But this can stay here for the time being.

poetry.lock
```diff
@@ -1,10 +1,15 @@
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
marcosertoli, line 1: how does this change when passing to Python 3.11?

I would recommend waiting for the 311 pull to be completed and then adapt this pull accordingly.

marcosertoli (commented):
I recommend waiting for the 311 pull request to be completed so you can merge those changes in and make sure it all works with the new release

indica/operators/spline_fit.py
pixelifytica: Makes sense, I only moved the import inside the function as `ReadST40` failed to import if (like me) you don't have MDSPlus installed, so this just scoped the import to not block importing from `spline_fit` in general

poetry.lock
pixelifytica: Yes I absolutely agree, with the move to 3.11 we should probably also review our dependencies and update the pinned versions, so that will lead to re-generating the `poetry.lock` if we do.

As `idlbridge` is a bit of a hacky package to install - and is basically abandoned anyway - I'm thinking I'll actually remove it from `pyproject.toml` and for the specific use case someone would want it they can install it manually.

pixelifytica:
Tested against the Python 3.11 changes and everything looks good to
go, read for review when you can.

I've removed `idlbridge` from the dependencies and trust if a user
needs it (since it's currently only for 1 thing in `PPFReader`) they
can install it separately, so no point in having it in the
dependencies.

Also added a small check to the abstract reader methods to check if a
`database_result` quantity is not `None` before assigning it. Some of
the JET systems only optionally produce some data so I figure it's
useful to have a catch for if a quantity is missing, and doesn't
affect the behaviour at all if data is present.

marcosertoli (approved):
Good from my end Evie.

For info and relevant to this pull, @michael-gemmell and I have been discussing a hefty refactoring of the readers. Main reasons for this is: 
- one has to edit 4 different files to write any new reader method. 
- Info on coordinates can be stored in the conf files and so much more can be abstracted. 
- Worth moving SAL-specific reader to a class e.g. mdsutils. 
- With all the above, we would only need 1 reader that takes in the reader class (e.g. mdsutils) and the machine-specific configuration file (e.g. st40conf).

He will go ahead and make a draft during August and we'll keep you and @hakosaj informed.

PR #328, opened 2024-06-20
Jussihakosalo/indica311

hakosaj:
Indica now compatible with 3.11.


indica/readers/abstractreader.py
```diff
@@ -269,17 +268,19 @@ def get_charge_exchange(
         """
         Reads Charge-exchange-spectroscopy data
         """
+
         database_results = self._get_charge_exchange(
             uid, instrument, revision, quantities
         )
-
+        print("charge4")
```
michael-gemmell, line 275: What are these for?

indica/utilities.py
```diff
@@ -320,14 +320,13 @@ def format_coord(data: LabeledArray, var_name: str):
...
-    coords = [(var_name, DataArray(data, dims=var_name))]
-    return format_dataarray(data, var_name, coords)
+    return format_dataarray(data, var_name, {var_name: data})
 
 
 def format_dataarray(
     data: LabeledArray,
     var_name: str,
-    coords: List[Tuple[str, Any]] = [],
+    coords: Dict[str, Any] = {},
```
michael-gemmell, line 329: This should be = None right? Since it's mutable

michael-gemmell (commented):
Looks good to me, just have to fix CI/CD.

indica/readers/abstractreader.py
hakosaj: Just weird debugging I forgot to remove

marcosertoli (approved):
All good, just need to sort CI

pixelifytica:
Why have we dropped support for python 3.9 with these changes?

pixelifytica:
From what I can see the only incompatible package is `ipython`, everything else seems still work with 3.9 (at least they're supported versions, I'm just running some tests now to make sure nothing odd happens), so we're safe to keep 3.9 support without changing the supported versions of the other dependencies.

Do we need `ipython` as a dependency? I can't find it being used anywhere

marcosertoli:
We don't need ipython, so it can be deleted.

I use it, but it's my personal development route, so no need to keep it in there.

Marco


pixelifytica:
Not a problem to move `ipython` to the dev dependencies group then rather than remove it, we can also set it to be conditional on python version so it's only installed if supported

PR #331, opened 2024-06-24
Adding example runs for diagnostic models

marcosertoli:


marcosertoli:
I also fixed a bug in the example he-like crystal model that will now work if you want to use it

PR #333, opened 2024-06-25
Marcosertoli/fix equilibrium r shift

marcosertoli:


indica/equilibrium.py
```diff
@@ -85,27 +87,24 @@ def __init__(
         R_new = self.psi.R + self.R_offset
         z_new = self.psi.z + self.z_offset
         self.psi = self.psi.interp(R=R_new, z=z_new)
-
-        self.faxs["R"] -= self.R_offset
-        self.faxs["z"] -= self.z_offset
```
michael-gemmell, line 90: Why are we not shifting faxs?

michael-gemmell (approved):

PR #338, opened 2024-07-10
Marcosertoli/zeff bremss reader

marcosertoli:
Added ZEFF_BREMS reader + minor mods for consistent release including:

- README changes
- Added @michael-gemmell  and @hakosaj  @pixelifytica in authors, maintainers --> please check!
- Temporarily commented out link to read the docs until it is fit for purpose!


README.md
```diff
@@ -26,18 +24,14 @@ ReadTheDocs](https://indica-ukaea.readthedocs.io/en/latest/), which is also stil
     - *pip install poetry==1.7*
     - *poetry update*
```
pixelifytica, line 25: I know it wasn't changed in this PR, but I only just noticed this says `poetry update`. Should be `poetry install` or `poetry install --sync`.

pixelifytica (approved):
If you could change the `poetry upgrade` bit in the README everything looks good to me otherwise. Make sure you merge after #320 since this includes the same commits.

indica/readers/abstractreader.py
```diff
@@ -15,20 +15,24 @@
...
-from indica.numpy_typing import OnlyArray
 from indica.numpy_typing import RevisionLike
 from indica.readers.available_quantities import AVAILABLE_QUANTITIES
 from indica.utilities import format_dataarray
 
 
 def instatiate_line_of_sight(
-    location: OnlyArray,
-    direction: OnlyArray,
+    database_results: dict,
```
michael-gemmell, line 24: Is there a reason for doing this? I don't like hiding a functions args in a dictionary; it makes it less clear what a function is doing.

michael-gemmell (changes requested):
Looks good to me

indica/readers/abstractreader.py
marcosertoli: I don't think it makes sense to use up 3 lines in every reader method when you can have 1. 
We could also move it to a private method to make it even clearer?

pixelifytica: Maybe a compromise option, could use the "**" expansion for function args? So keep the named arguments being the same as the keys in `database_results` and in the calls do `instantiate_line_of_sight(**database_results)` ? Would just need an extra `*args` and `**kwargs` in the `instantiate_line_of_sight` arguments to catch the extra values we're not interested in, then should work?

michael-gemmell (approved):

PR #342, opened 2024-07-17
Marcosertoli/separate examples

marcosertoli:
Making some order, separating examples from the class files.


hakosaj (approved):
Good stuff! Adds clarity nicely.

PR #343, opened 2024-07-24
Marcosertoli/fix time

marcosertoli:


marcosertoli:
Not sure why surf reading test fails in CI, since it works ok locally.
Trying it once more.

hakosaj (approved):
Always love more clarity!

PR #344, opened 2024-07-30
Marcosertoli/reinstate tomo asym

marcosertoli:
Reinstated tomo asymmetry to be applied at analysis of tangential cameras.
Fixed minor bugs and consistency.


indica/readers/st40reader.py
```diff
@@ -390,7 +391,8 @@ def _get_charge_exchange(
                     self.QUANTITIES_MDS[instrument][q],
                     results["revision"],
                 )
-            except TreeNODATA:
+            except Exception as e:
```
michael-gemmell, line 394: Generally it's better to catch specific exceptions rather than all of them. Should have a look at this when we rework the readers

michael-gemmell (approved):

indica/readers/st40reader.py
marcosertoli: I completely agree!

PR #349, opened 2024-08-01
R_shift had empty error coord and was causing problems for assigning the equilibrium to plasma

michael-gemmell:
R_shift had empty error coord and was causing problems for assigning the equilibrium to plasma

### DIFF
--- a/indica/readers/abstractreader.py
+++ b/indica/readers/abstractreader.py
@@ -219,6 +219,10 @@ class DataReader(BaseIO):
                 dims = ["t"]
             else:
                 raise ValueError(f"Unknown quantity: {quantity}")
+            if quantity == "R_shift":
+                include_error = False
+            else:
+                include_error = True
 
             data[quantity] = self.assign_dataarray(
                 instrument,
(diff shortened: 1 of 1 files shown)


indica/readers/abstractreader.py
```diff
@@ -219,14 +219,20 @@ def get_ppts(
                 dims = ["t"]
             else:
                 raise ValueError(f"Unknown quantity: {quantity}")
+            if quantity == "R_shift":
```
marcosertoli, line 222: I don't understand where this is throwing an error. In assign_dataarray, if there is not ".._error" quantity, then a default error = xr.zeros_like(data) is assigned. 

I haven't had any issues with this reader and have included it in both the Zeff and the Prad workflow.

If possible, I would sort this out where you get the error and leave this reader method as it was.

marcosertoli (commented):
I would fix where you get the error and revert to reader method as in the main branch.

indica/readers/abstractreader.py
michael-gemmell: The issue is that the R_shift gets a default error and then that propogates to the equilibrium which then gives me some error when I try plasma.set_equilibrium

marcosertoli: What error? 
I'd rather find out what's going wrong, eventually changing the R_shift assignment in the Equilibrium if you needed.

marcosertoli (changes requested):

PR #356, opened 2024-09-04
Mg/bayes opt with bo

michael-gemmell:
Bayesian Optimisation is now included in BDA!!!

Most of the changes are from previous PRs yet to be merged. Check workflows/optimiser_context for the BO implementation


indica/configs/workflows/bda_run/model/tws.yaml
marcosertoli, line 1: what are these setting files for?

indica/configs/workflows/bda_run/reader/mock.yaml
marcosertoli, line 1: How can you clarify what each file is for? Any way you can add comments.
I'm getting lost in the number of yaml files...

indica/converters/line_of_sight.py
```diff
@@ -75,7 +75,8 @@ def __init__(
     ):
 
         self.instrument_name: str = name
-        self.name = f"{name}_line_of_sight_transform"
+        self.type = "line_of_sight_transform"
```
marcosertoli, line 78: what do you need this for? the class name is self-explanatory and can be read within python if required

indica/models/plasma.py
```diff
@@ -837,7 +833,7 @@ def __call__(self):
 
 
 class PlasmaProfiles:
```
marcosertoli, line 835: hasn't this been superseded by the new class you've built?

connected to my comment on the pull request for this, I'd anyway save that new class in this file.

indica/models/helike_spectrometer.py
```diff
@@ -383,6 +383,7 @@ def __call__(
         moment_analysis: bool = False,
         background: float = None,
         pixel_offset: int = None,
+        norm_y: xr.DataArray = None,
```
marcosertoli, line 386: why put this in the model? this seems something you'd like to have in a workflow, leaving the model as light as possible...

indica/profilers.py
marcosertoli, line 1: as in comment in the other pull, I think we should make a new profiler/ folder, and separate the classes in this file in 3 different files (abstract + the child profilers)

indica/configs/workflows/bda_run/basic_run.yaml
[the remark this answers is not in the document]
michael-gemmell: It makes a lot of sense though I am worried that testing will be problematic.

indica/profilers.py
michael-gemmell: Seems like a good idea

indica/readers/abstractreader.py
```diff
@@ -219,14 +219,20 @@ def get_ppts(
                 dims = ["t"]
             else:
                 raise ValueError(f"Unknown quantity: {quantity}")
+            if quantity == "R_shift":
```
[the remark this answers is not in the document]
michael-gemmell: Did you see my comment on what was going wrong?

indica/models/helike_spectrometer.py
michael-gemmell: I was toying around with not fitting the argon density since it tends to cause a lot of headache (I'm undecided if this is needed). I don't see how to feasibly do this outside of the model.

indica/operators/atomic_data.py
```diff
@@ -536,6 +535,11 @@ def __call__(  # type: ignore
 
             self.F_z_t = F_z_t
         else:
+
+            # TODO: FIX THIS (Nh is ignored!?!)
+            # Not too mention why does F_z_t have extra dims
+            # (which are dropped..) other than Te / Ne / Nh???
```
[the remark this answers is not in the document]
michael-gemmell: I think I understand this F_z_t a bit more... As it is know Nh is fixed at each radial point which isn't great,

indica/converters/line_of_sight.py
michael-gemmell: I was imagining having some kind of dynamic behaviour based on the self.type, but that hasn't been useful yet.

indica/configs/workflows/bda_run/reader/mock.yaml
michael-gemmell: I can explain the hydra config philosophy. There are a few but it makes it easier to have a versatile workflow.

indica/configs/workflows/bda_run/model/tws.yaml
michael-gemmell: Some combinations of models for bda and their required settings

indica/configs/workflows/bda_run/basic_run.yaml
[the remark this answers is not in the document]
marcosertoli: We would test the Indica native stuff on Indica, and the dedicated ST40 workflows on the gitlab repo?

indica/operators/atomic_data.py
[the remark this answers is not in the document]
marcosertoli: I agree, but it's a good approximation. Better a rough Nh than none. 
We had previously discussed a tabulation of Fz on a Te, Nh matrix, but this is an overkill at the moment, since we're fixing Te and Ne using TS and Nh is an informed (wild) guess.

indica/models/helike_spectrometer.py
marcosertoli: I understand. 
But then wouldn't it be better to process the measured data e.g. subtract the background and normalize the w-line peak to 1?
Then you can impose background = 0 and the keyword norm_y could become simply normalize (or something similar) and simply measured_spectra/measured_spectra.max()?

indica/readers/abstractreader.py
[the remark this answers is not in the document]
marcosertoli: Yes, but didn't understand what exactly it was...

PR #357, opened 2024-10-31
removing all bda specific code

michael-gemmell:
Everything BDA specific has been moved to gitlab. https://tokamak-devlin.tokamak.local/gitlab/physics/diagnostics_analysis/bda


indica/configs/workflows/bda/plasma_profiler/default.yaml
marcosertoli, line 1: what is the difference between this and the config file indica/configs/workflows/bda/plasma/profiles/default.yaml?
why not put the profiler configs all in the indica/configs/profilers folder?

indica/profilers/profiler_spline.py
marcosertoli, line 1: Best to move the example in `if __name__ == "__main__"` to the examples folder

indica/workflows/bda/bda_driver.py
marcosertoli, line 1: Recommendation for another pull ;-) after we've refactored the readers: simplify the naming convention so PHANTOMS data-structure assignment can be simplified/abstracted/looped.

indica/workflows/bda/pca.py
marcosertoli, line 1: Do we not foresee any other potential application of this?
Is it worth keeping it, with minor mods, and moving it the indica/profilers/ folders for future use?

marcosertoli (approved):
All OK from my end, just a few minor comments that you might want to address before pressing the big green button :-)!

indica/configs/workflows/bda/plasma_profiler/default.yaml
michael-gemmell: I was trying to come up with a logical way to separate the plasma initialisation from the optimisation loop. This way I can create the plasma with one profiler and then optimise with another.

indica/workflows/bda/pca.py
michael-gemmell: It could be worth keeping but it needs some work before I am happy keeping it on the main branch. I may revisit it at some stage.

PR #363, opened 2024-12-02
Rewrite JET/SAL reader

pixelifytica:


indica/readers/salutils.py
```diff
@@ -35,23 +36,51 @@ class SALWarning(UserWarning):
...
         """Gets the signal for the given INSTRUMENT (DDA in JET), at the
         given revision."""
         path = self.get_sal_path(
-            uid,
-            instrument,
-            quantity,
-            self.get_revision(uid, instrument, quantity, revision),
+            uid=uid,
+            instrument=instrument,
+            revision=self.get_revision(uid, instrument, revision),
```
marcosertoli, line 82: This is unnecessary, since the get_revision method is already called in the _read_database method of the datareader.

marcosertoli (approved):
Looks good, please address only comment on revision and free to click the big green button :-).

Issue #366, opened 2024-12-18
Add capability to read Equilibrium from an EQDSK file


pixelifytica:
Suggest using the [FreeQDSK](https://github.com/freegs-plasma/FreeQDSK) package rather than write our own reader from scratch, then only needs a translation layer to get into an `Equilibrium` object (and, eventually, has tools to write to EQDSK files as well)

Issue #365, opened 2024-12-18
Add IMAS utils and a STEP reader


pixelifytica:

pixelifytica:
IMAS currently still requires a module from the ITER GitLab but progress is being made to full open source it (see [here](https://github.com/iterorganization/imas-python) and [here](https://github.com/iterorganization/imas-data-dictionary))

Issue #340, opened 2024-07-12
Add `UDA` reader for MAST-U data


pixelifytica:
Also separate `sal` reader back-end from `PPFReader` at the same time, `UDA` uses a similar structure to `sal` so will be similar work.

PR #367, opened 2024-12-18
Add extra readers

pixelifytica:
Work on closing #340 , #365 and #366 .


pixelifytica:
Added the `EQDSK` reader (#366) but also some extra model work, so figure this is worth merging in now rather than waiting for me to add the `UDA` and `IMAS` readers.

marcosertoli:
@pixelifytica , let me know once you've fixed the tests and I'll be happy to review.

pixelifytica:
That was a very weird one to fix, so the error came up with `test_interferometer` because it was technically 1D data as channel was a single value (0 in the case of the default object). `format_dataarray` tried to put channel as `[0]` which `DataArray` objects to, but setting channel as `0` in that case worked fine. I've added a catch in `build_dataarray` to turn 1-length arrays into just the value but I'm not convinced it's the best solution, @marcosertoli I'd like your opinion on this please.

indica/equilibrium.py
```diff
@@ -540,6 +543,62 @@ def enclosed_volume(
         volume = self.area.interp(rhop=rhop).interp(t=t)
         return volume, t
 
+    @classmethod
+    def read_geqdsk(cls, filename: Path) -> "Equilibrium":
```
marcosertoli, line 547: I would recommend to separate the reader form the class and add it to the /readers/ folder.
This is cleaner, you separate data reading from class instatiation, and you give the user the possibility to "just" read the eqdsk file and use if for something else. 

You will then need 2 lines to instatiate the Equilibrium class with this data, but it's worth it.

indica/models/polarimeter.py
marcosertoli, line 1: happy you've included this in! Will look into the details and send you some comments beyond the scope of this pull

marcosertoli (changes requested):
Hi Evie, minor changes of form more then substance!

indica/equilibrium.py
pixelifytica: Absolutely agree! Tried fitting it in to `DataReader` at first but figured it was too different to just read from a file, so it's just a standalone class

indica/datatypes.py
```diff
@@ -100,6 +100,7 @@
     "impurity": ("Impurity", "none"),
     "beamlet": ("Beamlet", "none"),
     "residence_time": (r"Impurity residence time $\tau$", "time"),
+    "magnetic_field": ("Magnetic field", "magnetic_field"),
```
marcosertoli, line 103: Is this the total magnetic field? I just noticed I've implemented toroidal, poloidal, radial and vertical, but not the total. 
I'd recommend keeping the standard used for the other entries, i.e.:
 "total_magnetic_field": ("$B_{tot}$", "magnetic_field")

indica/datatypes.py
```diff
@@ -119,6 +120,7 @@
         "Centrifugal asymmetry multiplier",
         "none",
     ),
+    "faraday_rotation_integrated": ("rad", "none"),
```
marcosertoli, line 123: The first item of the tuple is the DataArray LongName and the second is the Unit string identifier.
You'll have to add the Unit (e.g. "angle_radians" : "rad"), and you can pick a sensible LongName (e.g. "Faraday rotation").

indica/datatypes.py
```diff
@@ -196,4 +198,6 @@
     "absorbed_ohmic_power": ("$P_{ohm, abs}$", "power"),
     "safety_factor": ("q", "none"),
     "parallel_conductivity": (r"$\sigma$", "conductivity"),
+    "frequency": ("$omega$", "frequency"),
```
marcosertoli, line 201: What sort of frequency? Worth being more specific since the LongName comes out in the plot as well and explains what this variable really is

indica/equilibrium.py
```diff
@@ -77,8 +77,11 @@ def __init__(
 
         # Calculate rho-toroidal and 2D rho matrix
         self.rhot = np.sqrt(
-            (self.ftor - self.ftor.sel(rhop=0.0))
-            / (self.ftor.sel(rhop=1.0) - self.ftor.sel(rhop=0.0))
+            (self.ftor - self.ftor.sel(rhop=0.0, method="nearest"))
```
marcosertoli, line 80: Hadn't seen this during the previous iteration, there's something wrong and I think it's best to revert back to the main implementation.

The reason is that whatever equilibrium you read, it MUST have psin=0 and psin=1, hence it MUST have rhop=0 and rhop=1. 
If not, there must be something wrong with the equilibrium data...
Why did you have to implement this?

indica/readers/eqdskreader.py
marcosertoli, line 1: fantastic!

marcosertoli (changes requested):
Looks great Evie. Just a few minor points and one in the Equilibrium that needs looking into more closely.

indica/utilities.py
```diff
@@ -256,6 +258,8 @@ def build_dataarrays(
         coords: dict = {}
         for dim in dims:
             coords[dim] = data[dim]
+            if isinstance(coords[dim], Iterable) and len(coords[dim]) == 1:
```
marcosertoli, line 261: Resolved issue in pull request.

marcosertoli (changes requested):
Consider the pull request that works also for ST40.

marcosertoli (approved):

PR #368, opened 2024-12-20
Marcosertoli/fix examples

marcosertoli:
Fixed and added examples, including their high-level (run) tests to make sure they work in future releases


michael-gemmell (approved):
No notes. Looks good to me.

PR #379, opened 2025-02-26
Marcosertoli/bad channels

marcosertoli:


marcosertoli:
BAD_CHANNELS recently added to MDS+ to port from calibration to processed trees which avoids hard-wiring information in the analysis codes.

michael-gemmell (approved):
Seems fine. I'm not really sure what the utility of having "bad_channels" is vs setting those channels to NaN (within MDSPlus or post processing).

marcosertoli:
Both options were on the table. It was chosen to go for BAD channels for transparency & because of some erratic behaviour of the channels that can this way be checked post the post processing.

PR #381, opened 2025-03-05
Add CXRS concentrations and Zeff

pixelifytica:
Add concentration to CXRS model, otherwise mostly changes for JET reading to read concentrations and Zeff from CXRS. Needed to add `channel` to Zeff quantities since CXRS is multi-channel, but from testing the previous interferometer fix means this doesn't appear to affect reading diagnostics without channel


marcosertoli:
Hi @pixelifytica , happy to review once tests all pass :-) !

pixelifytica:
Sorry, I put this up and then sort of abandoned it because things suddenly happened! So I had a look and the tests seems to be failing because the BDA workflow uses C and He for a CXRS model but they're not in the plasma impurities. It passes fine if I add "c" and "he" to `impurities` in `indica/configs/workflows/bda/plasma/settings/default.yaml`, but @marcosertoli or @michael-gemmell could you comment on if that has repercussions for you?

michael-gemmell:
> Sorry, I put this up and then sort of abandoned it because things suddenly happened! So I had a look and the tests seems to be failing because the BDA workflow uses C and He for a CXRS model but they're not in the plasma impurities. It passes fine if I add "c" and "he" to `impurities` in `indica/configs/workflows/bda/plasma/settings/default.yaml`, but @marcosertoli or @michael-gemmell could you comment on if that has repercussions for you?

I can't think of any reason why this would be an issue. You have my blessing.

pixelifytica:
Hmm, tests should be passing now but it's "Missing download info for actions/cache@v2" to run the tests? Anybody know more about GitHub actions that can fix it? A quick search looks like actions/cache is up to v4 so maybe we just need to update it.

pixelifytica:
Bumping the version of actions/cache looks like it's fixed it

indica/models/charge_exchange_spectrometer.py
```diff
@@ -70,6 +72,11 @@ def __call__(
                 t = self.plasma.time_to_calculate
             Ti = self.plasma.ion_temperature.interp(t=t)
             Vtor = self.plasma.toroidal_rotation.interp(t=t)
+            Conc = (
+                self.plasma.ion_density.sel(element=self.element).interp(t=t)
+                * 100
```
marcosertoli, line 77: I'm torn between making the concentration a fraction or percent... 
I know I've defined the concentration as a percent in the datatypes, but this is inconsistent with the definition in the Plasma class which uses a fraction.

I'm leaning towards **fraction**, which makes it consistent with the Plasma where it is already actively used, and simplifies the problem.

What do you think?

marcosertoli (changes requested):
All good, we just need to decide a consistent definition for the **concentration**

indica/models/charge_exchange_spectrometer.py
michael-gemmell: Fraction is certainly better.

pixelifytica: Happy to go with fraction, agree the consistency with using the Plasma class is good. Do we leave it unit-less? Or add something (e.g. "nz/ne") so it's unambiguous that it's fraction rather than percent?

marcosertoli:
Super.

And unit-less will be fine.

marcosertoli (approved):

PR #382, opened 2025-03-14
Fix scale for JET CXRS concentrations

pixelifytica:
One small revision from the last PR that I forgot to change - JET CXRS concentrations come back as percent, so need to divide at the read step for consistency with the rest


pixelifytica:
It doesn't affect this PR but yep I saw the changes and have already rebased #401 and my new #409 on the new main, so I'll merge this and we can discuss more on those PRs!

PR #384, opened 2025-03-18
Marcosertoli/solps

marcosertoli:
SOLPS reader & adaptation of pinhole camera to deal with 2D (R, z) data - other models will have to be checked for this compatibility


indica/examples/example_diagnostic_models.py
```diff
@@ -22,7 +23,13 @@ def run_example_diagnostic_model(
     plasma.set_equilibrium(equilibrium)
     transform = transforms[instrument]
     transform.set_equilibrium(equilibrium)
-    model = model(instrument)
+
```
michael-gemmell, line 26: I think it makes more sense to have this in the example_model function.

michael-gemmell (changes requested):
No other notes

indica/examples/example_diagnostic_models.py
marcosertoli: Agreed and resolved

PR #386, opened 2025-03-20
passive spectrometer minor fix

michael-gemmell:
Small changes:

Fixing how kwargs are handled in example diagnostic models
Allowing PassiveSpectrometer to use ADF15 from specified charge states since there are gaps in OPEN-ADAS dataset


indica/examples/example_diagnostic_models.py
```diff
@@ -28,11 +28,6 @@ def run_example_diagnostic_model(
     transform = transforms[instrument]
     transform.set_equilibrium(equilibrium)
 
-    kwargs = {}
```
marcosertoli, line 31: if you don't have this, the pinhole camera model doesn't work - if you delete it from here, it should go in the example_ methods that call that model

indica/models/passive_spectrometer.py
```diff
@@ -17,12 +17,10 @@
 
 
 ADF15 = {
-    "h": dict(file_type="pju", year="96"),
-    "ne": dict(file_type="pju", year="96"),
-    "c": dict(file_type="pju", year="96"),
-    # "ar": dict(file_type="llu",
-    #            year="transport")
-    "ar": dict(file_type="ic", year="40"),
+    "he": dict(file_type="bnd", year="96", charges=[1]),
```
marcosertoli, line 20: thinking about this now, I have the feeling a few things could be moved around:
- ADF15 dictionary should probably go in a new adasconf.py in configs/readers/
- I'd reformat this ADF15 dictionary so that you can give a file per element & charge e.g. {"ne_0": dict(file_type="pju", year="96"}, "ne_8": dict(file_type="pju", year="96"}) --> you won't always use the same file_type and year for different ionisation stages!
- ADF11 dictionary in the adas reader should also be moved there
- pecs: dict in the passivespectrometer model init should be the ADF15 dictionary - the reformatting done internally
- following comment above: the read_and_format_adf15 method should be called in the passive spectrometer model - the parsing of the element and ionisation stages would have to be implemented.

marcosertoli (changes requested):

indica/models/passive_spectrometer.py
michael-gemmell: Good points on the ADF15 dictionary, I will implement it as you said. I disagree about putting the reading into the model, it makes more sense to me to decouple those parts. In general separating were datasets are created and used is good practice.

marcosertoli (approved):

PR #388, opened 2025-04-24
Crystal Calibration now correctly accounts for pixel binning.

michael-gemmell:
I added the theoretical estimate for the crystal calibration and the conversion from counts/s/nm to counts/s/pixel.

### DIFF
--- a/indica/models/helike_spectrometer.py
+++ b/indica/models/helike_spectrometer.py
@@ -29,13 +29,23 @@ class HelikeSpectrometer(AbstractDiagnostic):
     """
     Data and methods to model XRCS spectrometer measurements
 
+    calibration factor accounts for all geometric and optical losses between plasma source and detector (add ref).
+    N[counts/s] = 1/4pi * kappa * A_c * phy_y * R_c * sin(theta) * eta_lambda * integral(E_lambda dz)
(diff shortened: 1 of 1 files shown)


indica/models/helike_spectrometer.py
```diff
@@ -29,16 +29,13 @@ class HelikeSpectrometer(AbstractDiagnostic):
...
-    TODO: move window to call instead of init
     """
 
     def __init__(
         self,
         name: str,
         instrument_method="get_helike_spectroscopy",
-        etendue: float = 1.0,
-        calibration: float = 5.0e-19,
+        calibration: float = 1.5e-10,
```
marcosertoli, line 38: Worth adding a minor explanation in the inline docs on what this calibration means, e.g. that it includes both etendue and absolute calibration?

marcosertoli (approved):
Minor comment only - which probably applies to all the args and kwargs :-)! 

Probably for a different pull: are you planning to make this a child class of the Passive Spectrometer?

indica/models/helike_spectrometer.py
michael-gemmell: WIll do

PR #405, opened 2025-11-18
Merge JET workflow

pixelifytica:
Needs a lot of cleanup and refactoring before actually merging with main, but this gets the workflow I used for EPS off my JDC home directory at least! I don't know when I'll have the time yet to get this ready for merging, so will leave as a draft for now and keep rebasing onto new changes in main until then.


pixelifytica:
Combining with #425

PR #408, opened 2025-12-12
Aurora Operator

michael-gemmell:
Aurora is being included as an operator for impurity transport analysis. 

Can be used as a stand alone module with it's own inputs or as part of the Plasma class where it will determine the equilibrium fractional abundance of impurities given some defined diffusivity and convection coefficients.


michael-gemmell:
@marcosertoli and @hakosaj this PR is ready to be reviewed and merged whenever you guys have approved.

indica/examples/example_operators.py
```diff
@@ -362,3 +365,19 @@ def example_fit_ts(
...
+    Te = PLASMA.electron_temperature
+    Nh = PLASMA.neutral_density
+    D_z = PLASMA.diffusion_coefficient
+    V_z = PLASMA.convection_coefficient
+    operator = FractionalAbundanceAurora(
+        element="ar",
+        aurora_config=AuroraConfig,
+        equilibrium=EQUILIBRIUM,
+    )
+    fz_t = operator(Ne=ne, Te=Te, Nh=Nh, D_z=D_z, V_z=V_z, plot=plot)
```
marcosertoli, line 381: Since you need an Equilibrium, I'd avoid setting this in the config.

I see 2 options:
- set "t" in the operator call() similar to the diagnostic models - my preferred
- have a separate function generate the eqdsk file and input the file path in the call()

indica/examples/example_plasma.py
```diff
@@ -2,9 +2,15 @@
 
 from indica import Plasma
 from indica import PlasmaProfiler
+from indica.defaults.load_defaults import load_default_objects
 from indica.numpy_typing import Tuple
 from indica.profilers.profiler_gauss import initialise_gauss_profilers
 
+EQUILIBRIUM = load_default_objects(
```
marcosertoli, line 9: I'd rather leave this out so the user **must** consciously assign a geometry to the plasma. 

In the GitHub tests this will be the default object, but any other test can be adapted to have other gemetries.

indica/examples/example_plasma.py
```diff
@@ -72,7 +80,15 @@ def example_plasma(
             "impurity_density:ar.y0": nimp_y0[i],
             "impurity_density:ar.wcenter": nimp_wcenter[i],
         }
-
         plasma_profiler(parameters=parameters, t=t)
 
+    # Assign diffusion and convection coefficients
```
marcosertoli, line 85: This should be a DataArray with associated datatype etc. as every other plasma attribute

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+    ccd
+        charge exchange recombination coefficients string: f"ccd{year}_{element}.dat"
+    aurora_config
+        dictionary of Aurora namelist parameters, see configs/operators/aurora.py
+    """
+
+    def __init__(
+        self,
+        equilibrium: Equilibrium,
+        element: str = "ar",
```
marcosertoli, line 379: This is good and should be added to the old FractionalAbundance operator!

I'd recommend to follow the same variable names as in the Plasma for consistency and add also the main ion to the init i.e.:
- "element" --> "impurity" 
- "main_element" --> "main_ion"
You can specify in the in-line docs the change of variable name to maintain consistency in Indica.

And I'd also add a check that the self.nq is consistent with the atomic number returned by utilities.get_element_info!

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+        scd: str = None,
+        acd: str = None,
+        ccd: str = None,
+        aurora_config: dict = AuroraConfig,
+    ):
+
+        self.aurora_config = aurora_config
+        self.equilibrium = equilibrium
+        self.geqdsk = self.set_geqdsk()
```
marcosertoli, line 388: If you move "t" to the call, then this can be called there.

For typical impurity transport studies, D and v are assumed constant across the time of analyses --> you'll often have no need to change D and v, but will want to change the plasma parameters and the equilibrium

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+    """
+
+    def __init__(
+        self,
+        equilibrium: Equilibrium,
+        element: str = "ar",
+        scd: str = None,
+        acd: str = None,
+        ccd: str = None,
+        aurora_config: dict = AuroraConfig,
```
marcosertoli, line 383: similar to Fidasim, I would recommend setting to None any items in the Aurora config that should be set explicitly.

This includes:
- geqdsk_t_point - add to the call as previous comment
- imp & main_element - I'd add explicitly to the init (similar to the plasma)

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+        if D_z.ndim == 2:
+            _D_z = _D_z.transpose("rhop", "t").values
+            _V_z = _V_z.transpose("rhop", "t").values
+        elif D_z.ndim > 2:
+            raise Exception("D_z and V_z must be 1D or 2D.")
+
+        out = self.run_time_evolution(
+            D_z=_D_z, V_z=_V_z, plot=plot, times_DV=D_z.t.values
+        )
+        self.F_z_t = self.calc_fz(
```
marcosertoli, line 529: Do you need this anywhere else apart from in this operator?

If not, I'd either make it private (_calc_fz) or remove the function entirely and move the 3 lines of code from calc_fz to here directly - occupies same space, less jumping about.

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+                )
+
+        self.set_kinetic_profiles(
+            Te,
+            Ne,
+            Nh,
+        )
+        self.asim = aurora.aurora_sim(namelist=self.aurora_config, geqdsk=self.geqdsk)
+
+        # Interp DataArrays to Aurora rhop while converting to cm^-2 and cm^-1 units
```
marcosertoli, line 511: worth having a similar function as set_kinetic_profiles?

and why do you have to interpolate D and V on the aurora grid, and not do the same for Ne and Te?

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+    def set_geqdsk(
+        self,
+    ):
+        assert isinstance(self.equilibrium, Equilibrium)
+        geqdsk_filepath = self.equilibrium.write_to_geqdsk(
+            t_point=self.aurora_config["geqdsk_t_point"]
+        )
+        return omfit_eqdsk.OMFITgeqdsk(geqdsk_filepath)
+
+    def set_kinetic_profiles(
```
marcosertoli, line 413: private function "_set_..." used only internally to this class?

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+        # Interp DataArrays to Aurora rhop while converting to cm^-2 and cm^-1 units
+        _D_z = (
+            D_z.interp(rhop=self.asim.rhop_grid, kwargs={"fill_value": "extrapolate"})
+            * 1e4
+        )
+        _V_z = (
+            V_z.interp(rhop=self.asim.rhop_grid, kwargs={"fill_value": "extrapolate"})
+            * 1e2
+        )
+        if D_z.ndim == 2:
```
marcosertoli, line 520: This will depend on the order of the variables, right? best make the code check the order is how you want it?

indica/operators/atomic_data.py
```diff
@@ -349,6 +354,192 @@ def __call__(  # type: ignore
...
+        self.F_z_t = self.calc_fz(
+            out["nz"],
+            rhop=self.asim.rhop_grid,
+            time=self.asim.time_grid,
+            zimp=np.arange(self.asim.Z_imp + 1),
+            time_out=Te.t.values,
+            rhop_out=Te.rhop.values,
+        )
+
+        if plot:
```
marcosertoli, line 538: I'd take this out of the call and let the user call the function explicitly.

This function should be common to all fractional abundance operators as the F_z_t outputted should be the same!

indica/equilibrium.py
```diff
@@ -538,6 +548,61 @@ def enclosed_volume(
         volume = self.area.interp(rhop=rhop).interp(t=t)
         return volume, t
 
-    def write_to_geqdsk(self):
-        # TODO: Implement writing to geqdsk
-        raise NotImplementedError("Method not yet implemented")
+    def write_to_geqdsk(
```
marcosertoli, line 551: since you have this function, this convinces me even more that it's worth adding "t" to the __call__ of the fract abu operator and make the geqdsk on call!

indica/plasma.py
```diff
@@ -57,13 +60,16 @@ def __init__(
         full_run
             If True: compute ionisation balance at every iteration
             If False: calculate default and interpolate
+        aurora_run
```
marcosertoli, line 63: Please add a TODO so we can change this to a string defining the function name in the atomic_data.py, changing the old one to e.g. FractionalAbundanceCorona deleting any if/else statements

indica/plasma.py
```diff
@@ -426,6 +455,32 @@ def calc_fz(self):
                 self._fz[elem].loc[dict(t=t)] = fz_tmp.transpose()
         return self._fz
 
+    def aurora_fz(self, elements: tuple):
```
marcosertoli, line 458: This shouldn't be needed. 

Similar to the previous comment one should "just" call the required operator, the operator doing all the error-handling internally, and returning the same output.

Moving the common bits of the two operators to an abstract class will solve a lot of this

indica/plasma.py
```diff
@@ -607,13 +662,36 @@ def set_equilibrium(self, equilibrium: Equilibrium):
...
         """
-        Assigns default atomic fractional abundance and radiated power operators
+        Assigns default atomic fractional abundance and radiated power operators.
+        If self.aurora_run is True, uses aurora for ionisation balance calculation,
+        otherwise uses coronal approximation.
         """
-        fract_abu, power_loss_tot = default_atomic_data(
-            self.elements, full_run=self.full_run
-        )
+        if self.aurora_run:
```
marcosertoli, line 673: same to previous comment: call should only be modified to clarify which operator will be used without the need for if/else statements

indica/utilities.py
```diff
@@ -177,15 +177,21 @@ def coord_array(coord_vals: ArrayLike, coord_name: str):
...
-        "major_rad"
-        if coord_name == "R"
-        else "time"
-        if coord_name == "t"
-        else "norm_flux_pol"
-        if coord_name == "rhop"
-        else "norm_flux_tor"
-        if coord_name == "rhot"
-        else coord_name,
+        (
```
marcosertoli, line 180: is there a better way of coding this?

Can't even remember how we got to this...

pyproject.toml
```diff
@@ -25,12 +26,12 @@ dependencies = [
     "numpy==1.26.4",
     "mpmath>=1.3.0",
     "scipy>=1.9.2",
-    "xarray==2024.6.0",
+    "xarray>=2022.1.0",
```
marcosertoli, line 29: is this an issue for anything else?

and with the >=, will it just install the latest version? 

In general, should we rather specify the version of any package so we're sure it works the same for any user?

marcosertoli (changes requested):
A few minor comments! Happy to discuss if you need clarifications :-)

indica/examples/example_plasma.py
michael-gemmell: This is the case already

indica/utilities.py
michael-gemmell: Blame my linter, I haven't touched this.

pyproject.toml
michael-gemmell: We can, I needed an earlier version of Xarray for aurora

marcosertoli:
Worth raising the issue with whoever is managing Aurora?

michael-gemmell:
I've applied your comments @marcosertoli. Let me know what you think.

indica/operators/atomic_data.py
michael-gemmell: yes

indica/operators/atomic_data.py
```diff
@@ -349,6 +355,203 @@ def __call__(  # type: ignore
...
+    ):
+        # return self.asim.run_aurora_steady(D_z, V_z, plot=plot, **kwargs)
+        raise NotImplementedError
+
+    def run_time_evolution(
+        self, D_z: np.ndarray, V_z: np.ndarray, plot: bool = False, **kwargs
+    ):
+        return self.asim.run_aurora(D_z, V_z, plot=plot, **kwargs)
+
+    def calc_fz(
```
marcosertoli: This function can be deleted (it's called only once and only in this class) and line 467 moved to line 551

Also, best to change the variable name Nimp to something e.g. Nq because:
- This is resolved by ion charge, so is effectively the density of charge state "q"
- Nimp is used elsewhere in Indica and has the meaning of *total* impurity density, summed over all ionisation stages

indica/plasma.py
```diff
@@ -426,6 +456,32 @@ def calc_fz(self):
                 self._fz[elem].loc[dict(t=t)] = fz_tmp.transpose()
         return self._fz
 
+    def aurora_fz(self, elements: tuple):
+        # When using aurora, if either electron temperature or density are <= 0,
+        # skip calculation to avoid errors
```
marcosertoli, line 461: This should be necessary for whatever calculation!!!!

As in the previous comment, it seems to me that this is the only difference between *aurora_fz* and *coronal_fz*, so these extra functions might not be necessary?

Also, aurora and corona calculations should return an fz with the same dimensions, meaning that you shouldn't need to .transpose() one and not the other.

marcosertoli (changes requested):
Just a few minimal consistency changes, but otherwise all good to go!

indica/plasma.py
michael-gemmell: "Also, aurora and corona calculations should return an fz with the same dimensions, meaning that you shouldn't need to .transpose() one and not the other."
That would mean changing the coronal operator.

PR #422, opened 2026-03-02
Jussihakosalo/fidasim

hakosaj:
Please  ignore the 1300+ commits, I had to reorganise branch history a bit! But this is current main, and on top of that the fidasim-related changes. Pre commit fails because some comment lines are still too long but otherwise we good to go.


indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        self.einj = einj
+        self.pinj = pinj
+        self.current_fractions = current_fractions
+        self.ab = ab
+
+        self.plasma_ion_amu = plasma_ion_amu
+
+    def __call__(
+        self,
+        nbi_model: str = "FIDASIM",
```
marcosertoli: Move this to __init__

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        einj: float,
+        pinj: float,
+        current_fractions: List[float],
+        ab: float,
+        pulse: int = None,
```
michael-gemmell: No pulse info should be needed

indica/operators/nbioperator.py
michael-gemmell: As we discussed this should be in the __init__

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        ion_temperature: Optional[DataArray] = None,
+        electron_temperature: Optional[DataArray] = None,
+        electron_density: Optional[DataArray] = None,
+        neutral_density: Optional[DataArray] = None,
+        toroidal_rotation: Optional[DataArray] = None,
+        zeff: Optional[DataArray] = None,
+        *,
+        t: Union[DataArray, float, int],
+        pulse: int,
+        plasma: Optional[Plasma] = None,
```
michael-gemmell: This should not take the plasma as an input; only the profiles are needed. In addition I think the profiles should not be optional.

michael-gemmell: See how the other operator calls are used.

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        nbi_model: str = "FIDASIM",
+        ion_temperature: Optional[DataArray] = None,
+        electron_temperature: Optional[DataArray] = None,
+        electron_density: Optional[DataArray] = None,
+        neutral_density: Optional[DataArray] = None,
+        toroidal_rotation: Optional[DataArray] = None,
+        zeff: Optional[DataArray] = None,
+        *,
+        t: Union[DataArray, float, int],
+        pulse: int,
```
michael-gemmell: No pulse

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        self.electron_density = electron_density
+        self.neutral_density = neutral_density
+        self.toroidal_rotation = toroidal_rotation
+        self.zeff = zeff
+        self.t = t
+        if plasma is not None:
+            self.plasma = plasma
+        self.pulse = pulse
+
+        if self.plasma is not None:
```
michael-gemmell: This can go

indica/plasma.py
```diff
@@ -611,6 +610,8 @@ def build_atomic_data(self):
         """
         Assigns default atomic fractional abundance and radiated power operators
         """
+        from indica.operators.atomic_data import default_atomic_data
```
michael-gemmell, line 614: Should this be here?

indica/operators/st40_utils.py
```diff
@@ -0,0 +1,1170 @@
+from itertools import product
```
michael-gemmell: We can give this a more descriptive name like fidasim utils?

indica/operators/st40_utils.py
```diff
@@ -0,0 +1,1170 @@
...
+amu_kg = 1.67377e-27  # kg
+J_to_eV = 6.242e18
+
+# Isotope amu
+H_amu = 1.00794
+D_amu = 2.0141078
+T_amu = 3.0160492
+
+
+def extract_hda_plasma(
```
michael-gemmell: I don't think this belongs in utils.

indica/operators/beam_utils/fidasim_utils_plotting_legacy.py
```diff
@@ -0,0 +1,93 @@
+"""Archived plotting/debug snippets previously in fidasim_utils.py.
```
michael-gemmell: I don't recommend keeping archived code in the main repo.

examples/operators/nbiop_example.py
```diff
@@ -0,0 +1,53 @@
...
+    transforms = load_default_objects(machine, "geometry")
+    nbi_transform = transforms["tws_c"]  # Placeholder until dedicated NBI geometry.
+
+    equilibrium = load_default_objects(machine, "equilibrium")
+    plasma = load_default_objects(machine, "plasma")
+    plasma.set_equilibrium(equilibrium)
+    nbi_transform.set_equilibrium(equilibrium)
+
+    pulse = 13475
+    # fidasim_utils uses pulse when creating output dictionaries and paths.
```
michael-gemmell: We should change this to something more generic like a filepath

indica/operators/beam_utils/fidasim_utils.py
```diff
@@ -0,0 +1,657 @@
...
+        "nn": ctx["neutral_density"],
+        "ne": ctx["electron_density"],
+        "omegator": ctx["toroidal_rotation"],
+        "zeff": ctx["zeffective"],
+        "plasma_ion_amu": operator.plasma_ion_amu,
+    }
+
+    # Run TE-fidasim
+    run_fidasim_flag = True
+    sys.path.append(TE_FIDASIM_CODE_PATH)
```
michael-gemmell: Are we still using TE_FIDASIM?? Can we not get rid of it.

indica/operators/beam_utils/fidasim_utils.py
```diff
@@ -0,0 +1,657 @@
...
+
+    for beam in nbis:
+        if beam_id == beam["name"]:
+            fidasim.prefida(inputs, grid, beam, plasma, equil, fi_dist)
+
+    # If here then preprocessing was successful for this beam. Launch batch job.
+    # submit_fidasim_batch_job(beam_save_dir)
+
+
+# Might be used later.
```
michael-gemmell: Let's not keep unused code

michael-gemmell:
Have we figured out how we want to approach testing??

indica/examples/nbiop_example.py
```diff
@@ -0,0 +1,53 @@
...
+
+from indica.defaults.load_defaults import load_default_objects
+from indica.operators import nbioperator
+
+
+def run_nbi_operator_example():
+    """Build default inputs and run the NBI operator once."""
+    machine = "st40"
+    transforms = load_default_objects(machine, "geometry")
+    nbi_transform = transforms["tws_c"]  # Placeholder until dedicated NBI geometry.
```
marcosertoli: I see you're changing the spot and focal lengths later on, but this transform has 4 channels, so you might want to select 1 only to perform the tests.

examples/operators/nbiop_example.py
```diff
@@ -0,0 +1,53 @@
...
+    machine = "st40"
+    transforms = load_default_objects(machine, "geometry")
+    nbi_transform = transforms["tws_c"]  # Placeholder until dedicated NBI geometry.
+
+    equilibrium = load_default_objects(machine, "equilibrium")
+    plasma = load_default_objects(machine, "plasma")
+    plasma.set_equilibrium(equilibrium)
+    nbi_transform.set_equilibrium(equilibrium)
+
+    pulse = 13475
```
marcosertoli: No need for this if you're using only "phantom" data from the default objects

indica/examples/nbiop_example.py
```diff
@@ -0,0 +1,53 @@
...
+        einj=52.0,  # keV
+        pinj=0.5,  # MW
+        current_fractions=[0.5, 0.35, 0.15],
+        ab=2.014,
+        pulse=pulse,
+    )
+    nbi_op.set_transform(nbi_transform)
+    nbi_op.set_plasma(plasma)
+
+    neutrals_by_time = nbi_op(
```
marcosertoli: If you assign the plasma, then you don't need to set these - see further comment in the operator class

indica/operators/beam_utils/fidasim_utils.py
```diff
@@ -0,0 +1,657 @@
...
+        "te": ctx["electron_temperature"],
+        "nn": ctx["neutral_density"],
+        "ne": ctx["electron_density"],
+        "omegator": ctx["toroidal_rotation"],
+        "zeff": ctx["zeffective"],
+        "plasma_ion_amu": operator.plasma_ion_amu,
+    }
+
+    # Run TE-fidasim
+    run_fidasim_flag = True
```
marcosertoli: why do you need this? if you're calling a function called run_fidasim, you would expect to run it!

would it not be better to separate the run (this function) form the reading of the results (another one) so you can just read the data from a previous run if you want to and you can delete any run flag?

indica/operators/beam_utils/fidasim_utils.py
```diff
@@ -0,0 +1,657 @@
...
+                    data, dims=dims, attrs=dict(obj.attrs)
+                )
+
+        # Recursive visitor function
+        h5f.visititems(_visit)
+    # Build the actual dataset
+    return xr.Dataset(data_vars=data_vars, attrs=root_attrs)
+
+
+def _run_fidasim(operator, ctx: dict) -> dict:
```
marcosertoli: why is this a private method while prepare_fidasim is not?

indica/operators/beam_utils/fidasim_utils.py
marcosertoli: if this is to be used later, it would be best putting it somewhere else temporarily until you want to merge it in. 

otherwise, it's hundreds of commented lines

indica/operators/beam_utils/fidasim_utils_plotting_legacy.py
marcosertoli, line 1: please take this out of the pull and add it to the next iteration once it's fit for review :-)

indica/configs/operators/fidasim_configs.py
marcosertoli, line 1: should this go in the /config/ directory, under a new sub-directory /operators/?

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
+import numpy as np
+
+# Paths & environment: file locations for geometry,
+#  FIDASIM install, and default inputs/outputs.
+TE_FIDASIM_CODE_PATH = "/home/jussi.hakosalo/te-fidasim"
```
marcosertoli: I'd prefer to have a general config that is valid for everyone, using Path.home() and assuming where the installation will go, as well as a default path for the data to be stored in, in line with the ADAS reader so that, unless you give a path in the __init__ for where to find FIDASIM, it'll use the default one.

Also, the output should be in the same way as in the ADAS reader, e.g. in a folder called .indica/fidasim/ - again, see ADAS reader

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
...
+FIDASIM_BASE_DIR = "/home/jussi.hakosalo/fidasim/FIDASIM-2.0.0"
+FIDASIM_OUTPUT_DIR = "/home/jussi.hakosalo/fidasim_output"
+FIDASIM_BIN_PATH = FIDASIM_BASE_DIR + "/fidasim"
+TE_FIDASIM_FI_DIST_FILE = (
+    "/home/jussi.hakosalo/te-fidasim/9188_150_rfx/a5fidasim_distribution.h5"
+)
+NBI_USER = "jussi.hakosalo"
```
marcosertoli: no need for this, keep it general and use Path.home()

indica/configs/operators/fidasim_configs.py
```diff
@@ -0,0 +1,182 @@
...
+FIDASIM_BASE_DIR = "/home/jussi.hakosalo/fidasim/FIDASIM-2.0.0"
+FIDASIM_OUTPUT_DIR = "/home/jussi.hakosalo/fidasim_output"
+FIDASIM_BIN_PATH = FIDASIM_BASE_DIR + "/fidasim"
+TE_FIDASIM_FI_DIST_FILE = (
+    "/home/jussi.hakosalo/te-fidasim/9188_150_rfx/a5fidasim_distribution.h5"
+)
+NBI_USER = "jussi.hakosalo"
+
+# Simulation switches: enable/disable individual fidasim capabilities
+SIMULATION_SWITCHES = {
```
marcosertoli, line 16: these are all FIDASIM specific, right?

then best have a specific fidasim_config.py instead of a general nbi_configs

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
...
+        "time": time,
+        "runid": runid,
+        "comment": "test",
+        "result_dir": beam_save_dir,
+        "tables_file": fida_dir + "/tables/atomic_tables.h5",
+    }
+
+
+# Build NBI settings: beam energy/power/species mix used by FIDASIM.
+def build_nbi_settings(st40_beams):
```
marcosertoli: these functions should be in the fidasim_utils, not in the configs

marcosertoli: but do we need this at all???

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
...
+    return {
+        "einj": st40_beams["einj"],
+        "pinj": st40_beams["pinj"],
+        "current_fractions": np.array((st40_beams["current_fractions"])),
+        "ab": st40_beams["ab"],
+    }
+
+
+# Build plasma settings: ion mass and impurity charge state.
+def build_plasma_settings(plasma_ion_amu, imp_charge):
```
marcosertoli: same: do we need this at all??

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
...
+    return {
+        "ai": plasma_ion_amu,
+        "impurity_charge": imp_charge,
+    }
+
+
+# Beam geometry: RFX beamline configuration used to build the FIDASIM beam grid.
+def get_rfx_geo():
```
marcosertoli: this is not used at all - what do you need this for? 
Is it worth to move it out of this pull for the time being?

indica/operators/nbi_configs.py
```diff
@@ -0,0 +1,182 @@
...
+    # rfx["widy"] = 1.0
+    # rfx["widz"] = 1.0
+    # rfx["divy"] = np.array([0.001,0.001,0.001])
+    # rfx["divz"] = np.array([0.001,0.001,0.001])
+
+    return rfx
+
+
+# Beam geometry: HNBI beamline configuration used to build the FIDASIM beam grid.
+def get_hnbi_geo():
```
marcosertoli: same as above - keep it safe, but remove from pull

indica/operators/nbioperator.py
marcosertoli: no need

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+class NBIOperator(Operator):
+
+    """This operator should be operating on a standard plasma+profiles, and spit out
+    fast neutral density and fast particle pressure. I believe it does.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        einj: float,
```
marcosertoli: I'll propose some better naming for all these variables :-)!

indica/operators/nbioperator.py
marcosertoli: not here - plasma assigned through set_plasma

indica/operators/nbioperator.py
marcosertoli: See models/helike_spectrometer.py to copy style of call: 
if self.plasma --> use plasma!

then assign attributes to class

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+            if self.electron_density is None:
+                self.electron_density = self.plasma.electron_density
+            if self.neutral_density is None:
+                self.neutral_density = self.plasma.neutral_density
+            if self.toroidal_rotation is None:
+                self.toroidal_rotation = self.plasma.toroidal_rotation
+            if self.zeff is None:
+                self.zeff = self.plasma.zeff
+
+        if (
```
marcosertoli: This is required only if the inputs are given in the call. The plasma will have it all.

check Helike model for reference

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+            self.ion_temperature is None
+            or self.electron_temperature is None
+            or self.electron_density is None
+            or self.neutral_density is None
+            or self.toroidal_rotation is None
+            or self.zeff is None
+        ):
+            raise ValueError("Give inputs or assign plasma class!")
+
+        if self.t is None:
```
marcosertoli: best check all inputs at the beginning and altogether - Helike model is a good example

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        if self.transform is None:
+            raise ValueError("transform is required (set it before calling)")
+        if (
+            not hasattr(self.transform, "equilibrium")
+            or self.transform.equilibrium is None
+        ):
+            raise ValueError("transform is missing equilibrium data")
+
+        # Resolve which NBI model runner to use for this call.
+        model = nbi_model
```
marcosertoli: no need for double variable, right?

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+        model_key = str(model).strip().upper()
+        model_handler = self._get_model_handler(model_key)
```
marcosertoli: you can probably just call this ``self._get_model_handler(self.model.strip().upper())``

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+                eq.rhop.z,
+                t=time,
+            )
+            br = br.values
+            bz = bz.values
+            bt = bt.transpose("z", "R").values
+
+            # TODO - remove this. Currently using this bt estimate
+            #  as otherwise it wont run due to nan errors.
+            irod = 3.0 * 1e6
```
marcosertoli: this can be deleted since you're calculating Bt above ;-)

indica/operators/nbioperator.py
```diff
@@ -0,0 +1,221 @@
...
+                "neutral_density": neutral_density_t,
+                "toroidal_rotation": toroidal_rotation_t,
+                "zeffective": zeffective,
+            }
+            contexts.append(ctx)
+
+        return contexts
+
+    @staticmethod
+    def _get_model_handler(model_key: str):
```
marcosertoli: wouldn't it be better to put this in the __init__ where the configs are and import it at the top of the operator? 

see example MODELS_METHODS in /models/__init__.py

indica/operators/pi_spec_13475_t_0.090000.json
marcosertoli, line 1: same as other old files - keep them safe but take them away from this pull

indica/operators/st40_utils.py
marcosertoli, line 1: what is this???

indica/plasma.py
marcosertoli, line 614: why has this been moved here?

marcosertoli (changes requested):
Good general structure, just a few outstanding mods

indica/examples/nbiop_example.py
michael-gemmell: No plasma assigning is needed 🙅

indica/operators/beam_utils/fidasim_utils_plotting_legacy.py
hakosaj: Only temporary, will be merged in the next PR (after chatting with Hannah). Parts of this code will be used anyway, shoulda just left this file outta the PR I suppose

indica/examples/nbiop_example.py
hakosaj: Wait, I'm receiving conflicting info. Are we assigning or not?

indica/plasma.py
hakosaj: Lazy importing so we only import this when we call this function, thus making the import structure more robust. Prevents a certain circular import.

indica/operators/st40_utils.py
hakosaj: This is the util file I mentioned, that we spoke of, the long one TE_fidasim uses?  I just now refactored it out though.

indica/operators/beam_utils/fidasim_utils.py
hakosaj: We are not? This is simply a name of a path that is used to store the fidasim output/input files, we get this from the configs.

indica/plasma.py
hakosaj: Yes! Lazy importing to prevent a circular one in some cases.

indica/operators/nbioperator.py
hakosaj: Removed and redone paths

hakosaj: P sure operator needs these

indica/operators/nbi_configs.py
hakosaj: What do you mean it is not used? It is used to define the beam grid. Prepare fidasim line 320

indica/operators/nbi_configs.py
hakosaj: What do you mean it is not used? It is used to define the beam grid. Prepare fidasim line 320

indica/operators/nbi_configs.py
hakosaj: Used in prefida call

indica/operators/nbi_configs.py
hakosaj: This is used to construct the prefida inputs

examples/operators/nbiop_example.py
hakosaj: Not fully sure what you mean

indica/examples/nbiop_example.py
hakosaj: Needs clarification!

indica/operators/nbi_configs.py
hakosaj: Also, these are more configs than util functions! But we'll talk about this later today.

indica/examples/nbiop_example.py
hakosaj: Invent a transform with just one channel! Pick these from the HNBI nodes, and pick the focal lengths etc.

indica/operators/nbi_configs.py
hakosaj: These should be read in, fine to have defaults but also offer a way to read them in.

indica/operators/nbi_configs.py
hakosaj: Also MDS+ read instead of config! Can have a default but do offer a way to read it in. This is machine/device specific.

indica/operators/nbi_configs.py
hakosaj: Also MDS+ read instead of config! Can have a default but do offer a way to read it in. This is machine/device specific.

hakosaj: Should be fida config instead

indica/operators/nbi_configs.py
hakosaj: Fida specific and not NBI general

indica/operators/nbi_configs.py
hakosaj: Actually this is unnecessary and I should just use these from the object.

marcosertoli:
Hello @pixelifytica , how's the review of this code proceeding?

indica/configs/operators/fidasim_configs.py
```diff
@@ -0,0 +1,80 @@
+from pathlib import Path
+
+# Paths & environment: file locations for FIDASIM install and output.
+# FOR FIDASIM INSTALLATION:
+# https://d3denergetic.github.io/FIDASIM/page/01_getting_started/01_install.html
+# And just make sure the installation is stored in .Indica folder
+FIDASIM_ROOT = Path.home() / ".Indica" / "fidasim"
```
pixelifytica: This should be `.indica` (lowercase) for consistency with other parts of the code

indica/configs/operators/fidasim_configs.py
```diff
@@ -0,0 +1,80 @@
...
+FIDASIM_BASE_DIR = str(FIDASIM_ROOT / "FIDASIM-2.0.0")
```
pixelifytica: Since we require users to install FIDASIM independently I suggest we don't force it to be in the `.indica` directory, as they might use it else where.

Perhaps replace with:
```
FIDASIM_BASE_DIR = str(Path(getenv("FIDASIM_DIR", FIDASIM_ROOT / "FIDASIM-2.0.0")).expanduser().resolve())
```
to re-use the `FIDASIM_DIR` environment variable that's set when you follow their install instructions

indica/configs/operators/fidasim_configs.py
```diff
@@ -0,0 +1,80 @@
...
+FIDASIM_OUTPUT_DIR = str(FIDASIM_ROOT / "output")
+FIDASIM_BIN_PATH = str(Path(FIDASIM_BASE_DIR) / "fidasim")
+FIDASIM_FI_DIST_FILE = str(
```
pixelifytica: See comment about this in `fidasim_nbioperator.py`

indica/configs/operators/fidasim_configs.py
pixelifytica:
```suggestion
FIDASIM_BASE_DIR = str(Path(getenv("FIDASIM_DIR", FIDASIM_ROOT / "FIDASIM-2.0.0")).expanduser().resolve())
```

indica/operators/abstract_nbioperator.py
```diff
@@ -0,0 +1,152 @@
...
+        Zeff: DataArray,
+        MeanZ: DataArray,
+        target_element: str,
+        t: float,
+        file_name: Optional[str] = "",
+        pulse: int = 0,
+        machine: str = "tokamak",
+        prepare_kwargs: dict = {},
+        run_kwargs: dict = {},
+    ) -> dict:
```
pixelifytica, line 80: It wouldn't be essential but would be great to have these quantities be (optionally) picked up from an attached `Plasma` object similar to how diagnostic models work. Maybe some mix of profiles pulled from a plasma and other quantities are attributes set on the `NbiOperator` instance before calling? Happy to discuss and get everyone's thoughts

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+    "round": 2,
+    "circular": 2,
+}
+SHAPE_MAP_DEFAULT = 2
+
+
+class NbiFidasim(NbiOperator):
+    def prepare(
+        self,
+        fi_dist_file: str = FIDASIM_FI_DIST_FILE,
```
pixelifytica: Until we can generate a distribution file from InDiCA plasma state I don't think this should have a default value, that way the user knows they have to provide a distribution file rather than a more obscure file not found error otherwise

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+            self.neut_file
+        )
+        if reuse_existing_outputs and (not overwrite) and existing_outputs:
+            self._reuse_existing_outputs = True
+            return
+
+        self._reuse_existing_outputs = False
+        fidasim.prefida(inputs, grid, beam_cfg, plasma, equil, fi_dist)
+
+    def run(self, num_cores: int = 3, reuse_existing_outputs: bool = False):
```
pixelifytica: I think this should either be 1 or something related to `multiprocessing.cpu_count()` to be dynamic for the machine it's running on, 3 is an arbitrary number and there's a chance - however unlikely - that someone only has e.g. 2 cores

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+        print("Running beam code")
+
+        completed = subprocess.run(
+            [
+                FIDASIM_BIN_PATH,
+                self.fidasim_out,
+                f"{num_cores}",
+            ]
+        )
+        if completed.returncode != 0:
```
pixelifytica, line 299: Small bit of idiot proofing - if `returncode` is -11 it's _highly likely_ the user forgot to run `ulimit -s unlimited` before running. Worth adding a message to remind them? Speaking from personal experience here!

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+        nbi_settings = {
+            # Indica uses eV/W; FIDASIM input expects keV/MW.
+            "einj": self.energy * 1.0e-3,
+            "pinj": self.power * 1.0e-6,
+            "current_fractions": np.array(self.current_fractions),
+            "ab": self.nbi_element_info["A"],
+        }
+        plasma_settings = {
+            "ai": self.target_element_info["A"],
+            "impurity_charge": int(np.mean(self.MeanZ)),
```
pixelifytica: FIDASIM 2.0 only supports H (and isotopes), He3, B, and C for target impurities last time I checked, and will fail to find atomic data if another charge is used. For now I think impurity charge should be it's own input (either at `__init__` or as argument to `prepare`) and the user warned if choosing an unsupported impurity

pixelifytica (changes requested):
Sorry for the delay! Took a while to get back on top of things. Overall I think this is a fantastic implementation, thanks for all the work @hakosaj !

I've left a few comments in various places, only a couple of critical/breaking issues, mostly what I'd call snagging issues when trying to use this in a few different scenarios.

pixelifytica:
Not really a review: Before we merge we should have a quick chat about how to handle the commit history. My first thought would be to just re-commit the final changes onto main to effectively flatten things, or we could take a more nuanced approach if we'd like to keep the more granular history. Either way I'm worried about what the re-written history as it stands would do to main so let's discuss before merging.

pyproject.toml
pixelifytica, line 1: We need to add [h5py](https://pypi.org/project/h5py/) for `fidasim_nbioperator` and [scikit-image](https://pypi.org/project/scikit-image/) for FIDASIM's `utils.py` to work. Since FIDASIM is a bit of an optional extra tool we could put them in their own `fidasim` dependency group?

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+    "circular": 2,
+}
+SHAPE_MAP_DEFAULT = 2
+
+
+class NbiFidasim(NbiOperator):
+    def prepare(
+        self,
+        fi_dist_file: str = FIDASIM_FI_DIST_FILE,
+        save_dir: str = FIDASIM_OUTPUT_DIR,
```
pixelifytica, line 45: I'd also quite like to add `tables_file` as an input here, setting `FIDASIM_ATOMIC_TABLES = str(Path(FIDASIM_BASE_DIR) / "tables/atomic_tables.h5")` as the default in `fidasim_configs.py`. It's good to have the default if you don't want to touch it, but since you can build atomic tables with different settings would be good to allow users to choose a different file

indica/operators/fidasim_nbioperator.py
```diff
@@ -0,0 +1,836 @@
...
+    # Optional in-plane axis rotation.
+    if delta_ang != 0.0:
+        axis_angle = np.arctan2(beam_cfg["axis"][1], beam_cfg["axis"][0]) + delta_ang
+        beam_cfg["axis"] = np.array(
+            [np.cos(axis_angle), np.sin(axis_angle), beam_cfg["axis"][2]]
+        )
+        beam_cfg["axis"] = beam_cfg["axis"] / np.linalg.norm(beam_cfg["axis"])
+
+    # TODO: Check that these make sense!!!
+    rstart = transform._machine_dims[0][1] * 100.0
```
pixelifytica, line 879: I've been chasing an obscure bug where FIDASIM complains about generating neutrals already inside the plasma even though the beam clearly started outside. I thought this made sense originally but setting `rstart = np.sqrt(np.sum(transform.origin[0] ** 2)) * 100.0` (so it's based on the transform position rather than the machine bounds) seems to have fixed this. I don't fully understand what caused it yet, since it only happened with certain beam geometries I'm assuming something to do with the R-theta -> x-y grid conversions?

PR #426, opened 2026-05-07
Jh/noisemodel

hakosaj:
Note the currently I have added the new finalise_bkck call to just pinhole camera! 

I changed the abstractconvertor plotting a bit to support the newer xarray version.


indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
+from collections.abc import Callable
+
+import numpy as np
+
+
+def add_poisson_noise(data, typical_counts=100, reference=None, rng=None):
```
michael-gemmell: Probably this should not be under models, maybe operators makes more sense.

michael-gemmell (commented):
All of this looks pretty good to me. I might make some minor changes to method/variable names but the structure makes sense to me.

indica/models/noise.py
hakosaj: Makes sense, it's an easy thing to move. How do you feel about the way we are calling for the dictionary finalization at the end of the subclass model call?

indica/models/abstract_diagnostic.py
```diff
@@ -53,3 +54,49 @@ def __call__(self, *args, **kwargs) -> dict:
...
+            raise KeyError(
+                f"'{target_quantity}' not found in model output keys: "
+                f"{list(self.bckc.keys())}"
+            )
+
+        clean = self.bckc[target_quantity]
+        self.bckc[f"{target_quantity}_raw"] = clean
+        self.bckc[target_quantity] = noise_model(clean, **config)
+
+    def finalize_bckc(self, **kwargs) -> dict:
```
michael-gemmell: This should either be an abstract method that is created for each subclass or a method that is generally applicable for all subclasses in my opinion.

I would get rid of this and include apply_noise in the call. But I am open to other ways of doing it.

finialize_bckc can be it's own separate abstract method.

hakosaj: I agree. This is a method that is generally applicable for all the subclasses, is it not?  I wrote it with the idea that all of the subclasses could call it in an identical way.

The reason I don't want to do it in the call is that then the noise applying as the last step of the process would need to be custom define for each subclass. The way I did it here should require minimal changes to the subclasses!

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Add signal-dependent Poisson noise to continuous positive data.
+
+    Parameters
+    ----------
+    data : np.ndarray or xarray.DataArray
```
marcosertoli: Best stick with DataArrays only following our Indica standards. No need to include the possibility of np.ndarrays.

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+        Positive clean signal.
+    typical_counts : float
```
marcosertoli: What is this?

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+        Signal value corresponding to ``typical_counts``.
+        If None, use mean(data).
+    rng : np.random.Generator or None
+        Optional random generator.
+    """
+    if rng is None:
+        rng = np.random.default_rng()
+
+    input_data = data
+    data = np.asarray(input_data)
```
marcosertoli: Why do everything in np instead of DataArray?

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if np.any(data < 0):
```
marcosertoli: raise is a bit strong - I'd recommend applying it only to positive values and setting the random noise to 0 for negative data

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+        raise ValueError("Poisson noise requires non-negative data.")
+
+    if reference is None:
```
marcosertoli: I'm not sure I get this: is this e.g. the background level of the signal prior to plasma e.g. offset of a spectrometer camera? If so, then either you have it or you don't, i.e. `if reference is None` then reference = 0.

I'd actually change the keyword from `reference = None` to `reference = 0.`

Also, `reference` is a bit of a misleading variable name. Maybe something like background?

indica/models/noise.py
```diff
@@ -0,0 +1,64 @@
...
+        reference = np.mean(data)
+
+    if reference <= 0:
+        raise ValueError("reference must be positive.")
+
+    lam = typical_counts * data / reference
```
marcosertoli: I don't get this either...but I might just be rusty on my error propagation :-).

If we take photon noise, then the background detector noise is your `reference` and the photon noise that is Poisson distributed is `data - reference`, so:
`data_noise = rng.poisson(data - reference)`

If you have as well some background random noise, then you can define it as an additional quantity e.g. `background noise =  some random generator within a specified range`.

In general, a signal would be characterised by:
- background DC value & it's noise
- signal & signal noise

indica/models/abstract_diagnostic.py
```diff
@@ -53,3 +54,49 @@ def __call__(self, *args, **kwargs) -> dict:
             "{} does not implement a "
             "'__call__' method.".format(self.__class__.__name__)
         )
+
+    def apply_noise(self, noise: str, noise_config: dict | None = None):
```
marcosertoli: change `noise` to `noise_model` - no need have different variable names

indica/models/abstract_diagnostic.py
```diff
@@ -53,3 +54,49 @@ def __call__(self, *args, **kwargs) -> dict:
...
+                "noise_config={'target_quantity': 'brightness', ...}."
+            )
+        if target_quantity not in self.bckc:
+            raise KeyError(
+                f"'{target_quantity}' not found in model output keys: "
+                f"{list(self.bckc.keys())}"
+            )
+
+        clean = self.bckc[target_quantity]
+        self.bckc[f"{target_quantity}_raw"] = clean
```
marcosertoli: this is tricky naming... 
I'd rather make a new name e.g. `[f"{target_quantity}_clean"]` to leave `_raw` for raw data.

indica/models/abstract_diagnostic.py
marcosertoli: I'm not sure I understand fully either of your views...so I think I'll need to talk about it.

Briefly, my view.
From a functional perspective, the noise should be applied in the `_build_bckc_dictionary` method after the bckc dictionary has been built. So, you could delete this method entirely, move lines 98 and 99 in apply_noise and call this in each `_build_bckc_dictionary` method.

indica/models/noise.py
marcosertoli: as from other comments below, not sure about the input to this method.

Also, would be best to give typing hints, and rng can be a Callable with default what you're proposing below.

indica/models/pinhole_camera.py
```diff
@@ -147,7 +147,7 @@ def __call__(
 
         self._build_bckc_dictionary()
 
-        return self.bckc
+        return self.finalize_bckc(**kwargs)
```
marcosertoli: see comments above on calling apply_noise directly in the _build_bckc_dictionary method

marcosertoli (changes requested):
See what you make of my comments :-).

Also, @hakosaj , please check with Cary and Hannah (to start with) what they would need to characterize their signal noise!

indica/models/noise.py
hakosaj: Moved

indica/models/abstract_diagnostic.py
hakosaj: Moved and resolved. I did it the way Marco suggested.

indica/models/noise.py
hakosaj: Done

indica/models/noise.py
hakosaj: A poisson noise parameter that comments on "big" the noise is

indica/models/noise.py
hakosaj: Now we do DA

indica/models/noise.py
hakosaj: Done

indica/models/noise.py
hakosaj: Done

indica/models/abstract_diagnostic.py
```diff
@@ -53,3 +58,36 @@ def __call__(self, *args, **kwargs) -> dict:
...
+    def apply_noise(self, noise_model: str, noise_config: dict | None = None):
+        """
+        Apply noise to the back-calculated values.
+        The noise is applied to the quantity specified by
+        'target_quantity' in noise_config.
+        Noise models operate on DataArray values in self.bckc.
+        Preserves the original data in a new key with suffix
+          '_clean', e.g. 'brightness_clean' for 'brightness'.
+        """
+        if noise_config is None:
```
marcosertoli, line 67: Why not delete these 2 lines and simply put this in the method definition:
`def apply_noise(self, noise_model: str, noise_config: dict = {})`

indica/models/abstract_diagnostic.py
```diff
@@ -53,3 +58,36 @@ def __call__(self, *args, **kwargs) -> dict:
...
+            noise_config = {}
+        if not isinstance(noise_config, dict):
```
marcosertoli: This should also be captured during pre-commit and linting, so not really required.

indica/models/charge_exchange_spectrometer.py
```diff
@@ -95,7 +101,10 @@ def __call__(
         self.Vtor_at_channels = Vtor_at_channels
         self.Conc_at_channels = Conc_at_channels
 
-        self._build_bckc_dictionary()
+        self._build_bckc_dictionary(
```
marcosertoli: I think I'm missing something in this implementation: there's no information anywhere in each specific diagnostic model (or its abstract class) what `noise_model` and `noise_config` are.

Each diagnostic model must have it's own specific (default) noise model & config.

I'd recommend specifying this explicitly in either the __init__ of the model class (my preferred option) or in the __call__ (which would give you some flexibility in running it with different noise models, if desired).

But hidden in the kwargs is unclear.

indica/models/charge_exchange_spectrometer.py
```diff
@@ -95,7 +101,10 @@ def __call__(
...
+            noise_model=kwargs.get("noise_model", kwargs.get("noise")),
```
marcosertoli: not sure I get this:
`kwargs.get("attribute1", kwargs.get("attribute2")`

indica/models/charge_exchange_spectrometer.py
marcosertoli: A few reasons to put it in the __init__: 
- noise model and config can be stored as a attributes of the object - they are characteristics of the model!
- `_build_bckc_dictionary()` method can be called without any extra overhead of passing it any arguments because it will refer to the noise info internally e.g.

```
def _build_bckc_dictionary(self):
    self.bckc = {}
     ....
     self.bckc = build_dataarrays(bckc, self.quantities, transform=self.transform)
     ...
     if self.noise_model is not None:
         self.apply_noise(self.noise_model, self.noise_config)
```

indica/models/abstract_diagnostic.py
```diff
@@ -33,7 +34,11 @@ def set_parameters(self, **kwargs):
                 setattr(self, key, value)
 
     @abstractmethod
-    def _build_bckc_dictionary(self):
+    def _build_bckc_dictionary(
+        self,
+        noise_model: str | None = None,
```
marcosertoli: Why not the following:
`noise_model: str = None`
`noise_config: dict  = {}`

Is this bad practice that I don't know about?

indica/models/diode_filters.py
marcosertoli, line 1: comments in the charge_exchange_spectrometer model apply to all models

marcosertoli (changes requested):
Looks good, just a few minimal changes!

PR #425, opened 2026-04-13
Marcosertoli/new quantities

marcosertoli:

PR #427, opened 2026-07-06
JET config updates for workflow

pixelifytica:
@marcosertoli everything's a JET specific change except one line in `available_quantities.py`. I've been testing this branch today (which includes the #425 changes) and all seems to be working fine, so if you're happy to merge this I think #425 will basically be good to go soon. Will put an actually review on tomorrow so the machine is happy!


marcosertoli (commented):
Looks good to me @pixelifytica .

Are you going to also add a fix to the tests & change to UV in this pull?

pixelifytica:
@marcosertoli I'll fix it in #425 directly so we're working on the same place

PR #428, opened 2026-07-10
Marcosertoli/fix aurora

marcosertoli:
Found a fix for getting Aurora installed also using UV.

Needed moving to python 3.12 and old scipy release.

Mod in Readme explains how to do this.


marcosertoli:
@pixelifytica , @michael-gemmell seems like this fixes issues with Aurora, but please test how this works for you.
I've also updated the readme to be more representative of what Indica actually is.

marcosertoli:
I've also changed the rules for main to run 3.12 python tests instead of 3.11

pixelifytica:
Rather than suggesting `--python 3.12` for all the uv commands it supports a `.python-version` file so we should use that instead. @marcosertoli if you run `uv python pin 3.12` it'll make the file (or make it manually, it's just a simple text file) and commit that, then uv will use that version by default unless specifically overridden with the `--python` argument. Otherwise sounds good to me!

marcosertoli:
Good by me: can you run that change please and push to the same branch, Evie?

I've never used "pin".

Thanks,
Marco


pixelifytica (approved):

marcosertoli (commented):
Ok by me. 
@pixelifytica: can you Approve and merge? 
As pull author I am not allowed to approve ;-)

PR #432, opened 2026-07-29
Marcosertoli/refactor atomic data

marcosertoli:


marcosertoli:
@michael-gemmell @pixelifytica , the main scope of this pull is to refactor atomic data calculations so that we can install, run, test Indica without Aurora.

Additional things I've incorporated are:
- run most of the examples as tests to make sure they also work
- change Nh to Nn across the whole project
- delete un-used functions/operators (abstractoperator.py, strategies.py)

Let me know.
Marco

tests/integration/examples/test_examples.py
```diff
@@ -0,0 +1,43 @@
+import indica.examples.example_diagnostic_models as example_diagnostic_models
```
michael-gemmell: It seems strange that this is here as opposed to under /tests

indica/examples/example_equilibrium.py
```diff
@@ -5,7 +5,7 @@
 from indica.defaults.load_defaults import load_default_objects
 
 
-def flux_coords(machine: str = "st40"):
+def example_flux_coords(machine: str = "st40", plot: bool = False):
```
michael-gemmell: Personally I think it is redundant to have all the files and functions under the /examples dir to have 'examples' as a prefix, but obviously this is just an aesthetic choice.

michael-gemmell (approved):
Everything looks alright to me.

indica/operators/abstract_fractionalabundance.py
```diff
@@ -0,0 +1,143 @@
...
+CM, COLS = set_plot_colors()
+
+
+class FractionalAbundance(ABC):
+    """Abstract class to calculate ionisation balance using ADAS rate coefficients"""
+
+    def __init__(
+        self,
+        element: str,
+        scd_year: str = None,
```
pixelifytica, line 25: I'd prefer that we still pass the SCD/ACD/CCD data rather than just the years, so it doesn't force a specific ADAS reader instance. As an immediate example I use the reader with a path to our local ADAS data files rather than OpenADAS, but not forcing the reader keeps it modular in case we add other atomic data sources in the future. Happy to have it use a default reader to fetch the atomic data if not given, but I think we need to keep the option of providing the data directly.

indica/operators/powerloss_adas.py
```diff
@@ -0,0 +1,355 @@
...
+    prb
+        recombination and bremsstrahlung emissivity coefficients
+    prc
+        charge exchange emissivity coefficients
+    """
+
+    def __init__(
+        self,
+        element: str,
+        plt_year: str = None,
```
pixelifytica, line 30: Same argument here about passing the data directly rather than the years

indica/operators/powerloss_adas.py
```diff
@@ -0,0 +1,355 @@
...
+        _Nn *= Nn1
+        _Nn += Nn0
+        Nn.values = _Nn
+
+    _to_write = {"Te": np.array(Te), "Ne": np.array(Ne), "Nn": np.array(Nn)}
+
+    print("Calculate fractional abundance and cooling factors")
+    for elem in elements:
+        print(f"  {elem}")
+        fract_abu[elem] = FractionalAbundanceAdas(
```
pixelifytica: This is broken under the changed call signature of `FractionalAbundanceAdas`, would work if we changed it back however

pixelifytica (changes requested):
My only point is on passing data to Fzt and PL operators rather than years, otherwise this all makes sense and looks good to me

indica/examples/example_equilibrium.py
pixelifytica: Agreed, same as how we dropped the "get_" prefix from reader functions, because they're all in the examples directory and "example_*.py" files I think it's clear enough

tests/integration/examples/test_examples.py
marcosertoli: Agreed. Don't know why I left that there, give we already have tests.

I'll leave a few examples for other stuff that can be useful as snippets.

indica/examples/example_equilibrium.py
marcosertoli: agreed 100%, will change

indica/operators/abstract_fractionalabundance.py
marcosertoli: I share your view, @pixelifytica , and had changed this to make the init identical for ADAS and AURORA. 
To read the local ADAS data, you still use Indica's ADASReader but changing the path? If so, I see two possibilities:
- "just" adding the kwarg path_adas to specify where the data is to be found
- adding the option of giving the data instead of the years
Which would you both prefer @pixelifytica , @michael-gemmell ?

indica/operators/powerloss_adas.py
marcosertoli: see comment above

indica/operators/powerloss_adas.py
marcosertoli: Not tested anywhere and failed to spot it.

This is redundant given this is covered by default_atomic_data in defaults/generate_defaults.py, so I'd be tempted to delete function entirely.

indica/operators/abstract_fractionalabundance.py
pixelifytica: Aurora still needs the ADF11 data if I'm understanding it correctly, so the init is still the same for both if we pass the data directly? Adding a kwarg for adas_path works for my specific case but I still think we pass the data directly so any FractionalAbundance operator is dependent on the atomic data itself rather than the reader used. I don't expect we'll use something other than ADAS but it keeps the option open, especially if another machine wants to use their own atomic data library and writes a reader for it.

For how I'm using the reader, yep exactly! The ADASReader was originally written to either read from a local set of files if a path is given, or fetch from OpenADAS if no argument is given, so I often make use of the local ADAS install rather than pinging OpenADAS all the time.

marcosertoli: Aurora does not want the data, but the name of the adas file.

I share your perspective on avoiding reading data internally to the class wherever possible, but I also think we won't be running anything apart from ADAS.

Anyway, let me make a simple mod proposal to keep things as simple as possible.

pixelifytica: Oh I see where I've misread now! In which case I'd say the simplest thing is to also make reader an optional argument, where if `reader is None` we set `reader = ADASReader()`. Thoughts?

marcosertoli: Since we're working only with ADAS, let's leave it with the extra kwarg for the path for now.

If anyone wants to add other types of data, they'll modify accordingly and propose changes.

marcosertoli:
@pixelifytica , @michael-gemmell all changes done. Free to press the green button or send it back to me with more mods

pixelifytica (approved):
Looks good to me now!

PR #433, opened 2026-08-06
Add get_error method to database readers

pixelifytica:
@MetricBeckley has been working on the MAST-U/UDA reading capabilities and one thing that came up is the fetching of errors with the current reader system. It was always a bit hacky for JET anyway, and rather than come up with another workaround I figure we nip this in the bud.

The thought is having a `get_error` method that the datareader calls where each utils class can handle how to fetch errors (e.g. appending _err for mdsplus, I need to come up with a mapping to error dtypes for sal, etc...). I've added the method to mdsutils and mdsutils_mdsplus already to keep identical behavior to how it's done currently, so that should be an invisible change!

@marcosertoli and @michael-gemmell please let me know if you agree with handling errors like this going forwards and that it works fine for ST40 reading, @MetricBeckley would this fit well with the UDA reader?


pixelifytica:
@marcosertoli I've made the changes we talked about last week, `get_error` is now a machine reader specific thing with a default method in the abstract `datareader` to be overridden in subclasses. Please have a look and let me know your thoughts!

I'd particularly like your opinions on a check I removed in `build_dataarray`: It was replacing errors with NaNs where they were <0 (sensible, delta errors can't be <0) but also where the error was larger than the data (`_error / _data < 1`). Personally I disagree with that, if your errors are larger than your data then your data isn't good but it doesn't make the error invalid, and can even mask the fact there's bad data by hiding huge errors! So I've removed that part, thoughts?

